PR #314, opened 2026-03-21
 migrate Comment model to TypeScript and expand schema (Issue 7.21)

shubham-01-star:
**Summary**
Migrates the legacy `CommentModel.js` to strongly-typed TypeScript in `quotevote-next`.

**Changes**
- **Types**: Added `deleted`, made `postId` optional, and `startWordIndex`/`endWordIndex` required in `common.ts` and `mongoose.ts` to match the legacy schema.
- **Model**: Expanded `app/data/models/Comment.ts` with the full Mongoose schema, including text indexes and statics for `findByPostId` / `findByUserId`.
- **Tests**: Added 4 unit tests in `Comment.test.ts` to verify schema validation and static queries. 

All 405 tests, `tsc --noEmit`, and `pnpm lint` pass successfully.


Om7035 (changes requested):
quotevote-backend/app/data/models/Comment.ts
Inline comment A (on/near findByPostId static)
findByPostId(postId: string) should validate/cast ObjectId input before querying.
Suggest using Types.ObjectId.isValid(postId) and converting to new Types.ObjectId(postId) to avoid malformed query inputs and inconsistent behavior.

Inline comment B (on/near findByUserId static)
Same concern here: please validate/cast userId to ObjectId before querying.
Returning [] for invalid IDs is safer than hitting Mongo with invalid types.

Inline comment C (on/near statics + sort/filter query shape)
These statics filter by deleted != true and sort by created.
Please add compound indexes for these paths to avoid collection scans under load (e.g. { postId: 1, deleted: 1, created: 1 } and { userId: 1, deleted: 1, created: -1 }).

Inline comment D (on/near text index)
Text index on comments is useful, but please confirm this is required now and measured.
Text indexes increase write/index maintenance cost and should be intentional with query usage.

quotevote-backend/app/types/common.ts
Inline comment E (on/near postId?: string)
postId became optional — can we confirm this matches product rules?
If comments can exist without post linkage, service/API validators and consumers need explicit handling; otherwise this should remain required.

Inline comment F (on/near startWordIndex / endWordIndex required)
Making these required is a good consistency move, but please confirm all creation flows now populate both values.
Add/update tests for any API paths creating comments to prevent runtime regressions.

quotevote-backend/app/types/mongoose.ts
Inline comment G (on/near CommentDocument typing changes)
The type now omits and redefines fields (created, optional postId).
Please ensure this stays aligned with actual mongoose schema/runtime docs to avoid TS saying “safe” while runtime diverges.

quotevote-backend/__tests__/unit/models/Comment.test.ts
Inline comment H (on/near beforeAll setup comment)
Current setup comment suggests uncertainty around DB initialization.
For model statics/schema changes, please add deterministic integration tests (mongodb-memory-server) rather than relying only on mocks/spies.

Inline comment I (on/near mocked static tests)
These tests validate query construction, which is useful, but they don’t prove schema/index behavior.
Please add integration assertions for soft-delete filtering, sort order, and invalid ObjectId handling.

quotevote-backend/package.json (and lockfile)
Inline comment J (on Stripe dependency/type changes)
Stripe dependency/type updates appear unrelated to Comment migration scope.
Please move to a separate PR unless there is a strict dependency reason and it’s documented in this PR description.

PR #389, opened 2026-06-17
feat(guest-access): allow guests to browse read-only dashboard routes (RC1-003)

niamao:
## Summary
- Allow unauthenticated guests to access read-only dashboard routes: explore, post pages, and public profiles (`/dashboard/profile/:username`)
- Keep auth-required routes (settings, notifications, own profile shell, etc.) redirecting to login with `callbackUrl`
- Update landing hero search and activity/post navigation so guests reach content instead of being sent to login

## Test plan
- [ ] Visit `/dashboard/explore` while logged out — page loads without redirect
- [ ] Open a post URL while logged out — post page loads
- [ ] Open `/dashboard/profile/:username` while logged out — public profile loads
- [ ] Visit `/dashboard/settings` while logged out — redirects to `/auths/login?callbackUrl=...`
- [ ] Landing page hero search: post result → post page, creator result → profile, "See all" → explore with query
- [ ] `pnpm test -- src/__tests__/foundation/middleware.test.ts src/__tests__/utils/dashboard-routes.test.ts` passes

### DIFF
--- a/quotevote-frontend/middleware.ts
+++ b/quotevote-frontend/middleware.ts
@@ -1,5 +1,8 @@
 import { NextResponse } from 'next/server';
 import type { NextRequest } from 'next/server';
+import {
+  isGuestReadableDashboardRoute,
+} from './src/lib/dashboard-routes';
 
 // Routes that authenticated users should be redirected away from
 const AUTH_ROUTES = ['/auths/login', '/auths/signup', '/auths/request-access', '/auths/forgot-password'];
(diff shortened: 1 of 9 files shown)


motirebuma:
- This is a solid, well-thought-out PR and I'm happy to approve it. The core implementation is clean — dashboard-routes.ts is a nice separation of concerns, the middleware logic is correct, and the regex for public profiles does exactly the right thing by allowing /dashboard/profile/:username while blocking the bare /dashboard/profile shell. The test coverage is thorough too, especially the new dashboard-routes.test.ts and the updated middleware assertions.                                                                                 
                                                                                                                                                
- The landing page hero search improvement is a welcome UX win. Guests being sent to actual content instead of a login wall makes the app feel    
  much more welcoming to first-time visitors, and the activity list and PostCard guard removals are the right call now that those routes are    
  genuinely public.

A couple of small things worth a look before merging. Both ActivityList.tsx and PaginatedActivityList.tsx still import useGuestGuard even though all the ensureAuth() calls were removed — if that was the only usage in those files, lint will complain. Also, inside the control-panel block in middleware there's const payload = token ? decodeJwtPayload(token) : null — the token ? guard is technically unreachable there since any guest would've been redirected before hitting that line, so it reads as a little confusing even if it's harmless. Finally, creator.username ||
  creator._id as a profile URL slug is a safe fallback in theory, but routing someone to /dashboard/profile/<mongo-id> would land them on a 404 in practice, so it might be worth handling that case more gracefully.

None of these are blockers — the main feature works correctly and the tests back it up.

Approved.

PR #390, opened 2026-06-17
feat(auth): open login modal for protected actions instead of redirect (RC1-001)

niamao:
## Summary
- Replace hard redirects on auth-gated actions with a shared auth gate that opens `RequestInviteDialog` (login + request invite tabs) while keeping the user on the current page
- Mount `AuthGateDialog` globally; wire `useGuestGuard`, dashboard layout actions, Apollo `UNAUTHENTICATED` handling, and profile Message button to the modal flow
- Preserve `callbackUrl` on login so users can resume after authenticating

**Depends on:** #389 (RC1-003 guest read access) — merge RC1-003 first; this PR stacks on those commits until then.

## Test plan
- [ ] Logged out on `/dashboard/explore`: click Create / Messages / Notifications → modal opens, no redirect
- [ ] Logged out on a post: vote / comment / bookmark → modal opens
- [ ] Logged out on public profile: Message → modal opens
- [ ] Modal login tab → successful login returns to originating route via `callbackUrl`
- [ ] `pnpm test -- src/__tests__/components/RequestInviteDialog.test.tsx src/__tests__/context/AuthModalContext.test.tsx src/__tests__/components/Profile/ProfileHeaderMessage.test.tsx src/__tests__/utils/auth.test.ts` passes

### DIFF
--- a/quotevote-frontend/middleware.ts
+++ b/quotevote-frontend/middleware.ts
@@ -1,5 +1,8 @@
 import { NextResponse } from 'next/server';
 import type { NextRequest } from 'next/server';
+import {
+  isGuestReadableDashboardRoute,
+} from './src/lib/dashboard-routes';
 
 // Routes that authenticated users should be redirected away from
 const AUTH_ROUTES = ['/auths/login', '/auths/signup', '/auths/request-access', '/auths/forgot-password'];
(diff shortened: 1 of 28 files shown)


motirebuma (approved):
This change works. Glad to give it a go. A single file at the module level handles access checks quietly behind the scenes - linking old-style code like Apollo errors and login requirements into one smooth flow with React's interface system. Instead of weaving through layers just to pass down references, this sidesteps clutter neatly. One spot mounts the dialog early in the app structure; setup and removal happen cleanly when things shift around on load. Wrapping functions carefully keeps object identities steady where needed, while tracking what users see now gets handled better too

Swapping out window.location.href = '/auths/login' in apollo-client.ts for triggerAuth({ view: 'login' }) keeps people right where they are instead of yanking them away. A popup means logging in without losing your spot - cleaner, quieter, less jarring. Over in lib/utils/auth.ts, tossing the rigid https://quote.vote/invite?from= link was overdue; better late than stuck. Bringing callbackUrl into the login flow adds flexibility, which matters more than it sounds. That startsWith('/') check helps block shady external jumps - but watch out, because //evil.com sneaks past since slashes count. Adding !callbackUrl.startsWith('//') seals that gap without fuss.

One small thing stands out. That Suspense wrapping equestInviteDialog within AuthGateDialog feels unclear - maybe drop a line explaining why, especially if it ties to lazy loading or a suspending hook; someone later might wonder. Then there is requireAuthForAction, popping up in dashboard/layout.tsx, hooked just to Create here, yet the PR mentions Messages and Notifications too on mobile - better make sure those actually trigger the same gate before calling it done.

Still, auth-gate.ts lacks direct unit tests, even if parts of it run during broader checks. Because it plays such a key role here, trying out just one test for registerAuthGateHandler plus triggerAuthGate might help trust grow slowly. Yet none of that stops progress - what's built works right, especially when bigger pieces get tested well

Approved

Issue #384, opened 2026-06-15
RC1-032 Ensure highlight popup stays within viewport


flyblackbox:
# RC1-032 Ensure highlight popup stays within viewport

## Summary
The highlight popup should overlay the page and remain fully visible. It should not be cut off to the left or right at viewport edges.

## Area
Highlight Popup

## Current Behavior
Popup can be cut off by the left or right edge of the viewport.

## Expected Behavior
Popup is positioned and clamped within viewport bounds on desktop and mobile.

## Tasks
- [ ] Update popup positioning/clamping logic.
- [ ] Detect viewport edges and adjust placement.
- [ ] Consider portal rendering to avoid parent overflow clipping.
- [ ] Test text selection near all page edges.

## Acceptance Criteria
- Popup is positioned and clamped within viewport bounds on desktop and mobile.
- Behavior is consistent across relevant desktop and mobile views.
- No regressions are introduced in adjacent workflows.

## Test Instructions
1. Highlight text near left edge, right edge, top, and bottom.
2. Confirm popup remains fully visible.
3. Test desktop and mobile viewports.
4. Confirm popup overlays page content without clipping.

## Labels
`bug` `rc1` `highlight-popup` `layout` `frontend`

Issue #382, opened 2026-06-15
RC1-030 Prevent highlighted text from expanding on popup hover


flyblackbox:
# RC1-030 Prevent highlighted text from expanding on popup hover

## Summary
Highlighted text should remain stable when the user hovers over the highlight popup.

## Area
Highlight Popup

## Current Behavior
Highlighted text expands unexpectedly when hovering over the popup.

## Expected Behavior
Hovering over the popup does not alter the selected/highlighted text range or layout.

## Tasks
- [ ] Inspect hover handlers around highlighted text and popup.
- [ ] Remove layout-changing hover behavior from selected text.
- [ ] Ensure popup hover state affects only the popup controls.
- [ ] Test long highlights and wrapped lines.

## Acceptance Criteria
- Hovering over the popup does not alter the selected/highlighted text range or layout.
- Behavior is consistent across relevant desktop and mobile views.
- No regressions are introduced in adjacent workflows.

## Test Instructions
1. Highlight text on a post.
2. Move cursor over popup controls.
3. Confirm highlighted text does not expand or shift.
4. Repeat with multi-line highlighted text.

## Labels
`bug` `rc1` `highlight-popup` `ui` `frontend`

PR #401, opened 2026-06-25
Bugfix/issue 384 viewport clamping

OtavioXimarelli:
## PR — RC1-032: Render highlight popup in portal and clamp within viewport

**Branch:** `bugfix/issue-384-viewport-clamping` → `main`
**Closes:** #384
**Labels:** `bug` `rc1` `highlight-popup` `layout` `frontend`

---

### Problem

The highlight popup was rendered as a regular child element inside `SelectionPopover`'s DOM parent. This caused two distinct bugs:

**1. Overflow clipping**
Any ancestor with `overflow: hidden` (present in multiple layout containers) clipped the popup at the parent's edge. Selecting text near a container boundary made the popup appear cut off or invisible.

**2. No viewport clamping**
The old `computePopoverBox` positioned the popup relative to the `[data-selectable]` element's bounding box without checking viewport edges. Selecting text near the left or right edge caused the popup to render partially or fully off-screen.

The old positioning used three separate layout branches (desktop / tablet right-side / mobile) with coordinates relative to the parent container, all of which could still overflow the viewport:

```ts
// Old: position relative to parent container — could overflow viewport
setPopoverBox({
  top: selectionBox.top - popupH - 8 - targetBox.top,
  left: selectionBox.left - targetBox.left + selectionBox.width / 2 - popupW / 2,
  right: 0,
})
```

There was also a `handleMouseEnter` callback (introduced in the same branch) that mutated the live selection range on hover — a regression against the no-regressions acceptance criterion.

---

### Fix

**`SelectionPopover.tsx`** — four concrete changes:

**1. SSR guard (`mounted` state)**

Added a `mounted` state flag set in a `useEffect`. The component returns `null` on the server to avoid `document` access during SSR:

```ts
const [mounted, setMounted] = useState(false)
useEffect(() => { setMounted(true) }, [])
if (!mounted) return null
```

**2. Portal rendering**

The popup div is now rendered via `createPortal(…, document.body)`, escaping all ancestor overflow constraints. It overlays the full page regardless of the DOM hierarchy:

```ts
return createPortal(
  <div id="selectionPopover" ref={popoverRef} style={...}>,
  document.body
)
```

**3. Unified viewport-aware positioning with horizontal clamping and vertical flip**

Replaced the three-branch layout with a single algorithm that works at all viewport widths:

```ts
// Horizontal: centre over selection, clamp within [10px, viewport - popupWidth - 10px]
const viewportLeft = selectionBox.left + selectionBox.width / 2 - popoverWidth / 2
const clampedViewportLeft = Math.max(
  margin,
  Math.min(window.innerWidth - popoverWidth - margin, viewportLeft)
)
const pageLeft = clampedViewportLeft + window.scrollX

// Vertical: flip below the selection when there is not enough space above
if (selectionBox.top < popoverHeight + topOffset + 10) {
  pageTop = selectionBox.bottom + window.scrollY + topOffset  // below
} else {
  pageTop = selectionBox.top + window.scrollY - popoverHeight - topOffset  // above
}
```

Coordinates are now in page-absolute space (`viewport coord + scrollX/Y`) so the absolutely-positioned portal element tracks the selection correctly even on scrolled pages.

**4. Resize and scroll recalculation**

While the popup is visible, `computePopoverBox` is re-run via `requestAnimationFrame` on the next two frames (to allow the portal to paint and report its measured size), and on every `resize` and passive `scroll` event:

```ts
useEffect(() => {
  if (showPopover) {
    const rafId1 = requestAnimationFrame(computePopoverBox)
    const rafId2 = requestAnimationFrame(() => requestAnimationFrame(computePopoverBox))
    window.addEventListener('resize', computePopoverBox)
    window.addEventListener('scroll', computePopoverBox, { passive: true })
    return () => {
      cancelAnimationFrame(rafId1)
      cancelAnimationFrame(rafId2)
      window.removeEventListener('resize', computePopoverBox)
      window.removeEventListener('scroll', computePopoverBox)
    }
  }
}, [showPopover, computePopoverBox])
```

**5. Removed `handleMouseEnter`**

Removed the `handleMouseEnter` callback and `onMouseEnter` prop that were introduced in the first commit on this branch. The callback mutated the active `Range` on hover, which is the root cause of #382 and a direct regression against this issue's acceptance criterion.

**`SelectionPopover.test.tsx`** — two targeted changes:

- `container.querySelector('#selectionPopover')` → `document.querySelector('#selectionPopover')` throughout, because the popup now renders into `document.body` via the portal and is not a child of the React test container.
- Replaced `'handles mouse enter to expand selection'` (which asserted the buggy behavior with `toBeDefined()`) with `'does not alter selection or call onSelect when mouse enters the popover'` asserting `expect(onSelect).not.toHaveBeenCalled()`.

---

### Test results

```
PASS src/__tests__/components/VotingComponents/SelectionPopover.test.tsx
  SelectionPopover
    ✓ renders children when showPopover is true
    ✓ does not render children when showPopover is false
    ✓ calls onDeselect when selection is cleared
    ✓ applies custom topOffset
    ✓ renders above the post sticky action bar (z-index > 10) by default
    ✓ applies custom styles
    ✓ handles selection change events
    ✓ handles mobile selection events
    ✓ clears selection when showPopover becomes false
    ✓ does not alter selection or call onSelect when mouse enters the popover
    ✓ computes popover position correctly for desktop
    ✓ computes popover position correctly for tablet
    ✓ computes popover position correctly for mobile
    ✓ handles missing selectable element gracefully
    ✓ handles collapsed selection
    ✓ cleans up interval on unmount

Tests: 16 passed, 16 total
```

---

### Acceptance criteria

- [x] Popup is positioned and clamped within viewport bounds at all screen widths
- [x] Popup never clips against a parent container — rendered via `document.body` portal
- [x] Popup flips below the selection when space above is insufficient
- [x] Popup coordinates stay correct on scrolled pages (`scrollX` / `scrollY` offsets applied)
- [x] Position recalculates on `resize` and `scroll` while visible
- [x] SSR-safe — returns `null` until mounted on the client
- [x] Hovering the popup does not alter the active selection (no regression from #382)
- [x] No regressions in adjacent workflows

### DIFF
--- a/quotevote-frontend/src/__tests__/components/VotingComponents/SelectionPopover.test.tsx
+++ b/quotevote-frontend/src/__tests__/components/VotingComponents/SelectionPopover.test.tsx
@@ -189,7 +189,7 @@ describe('SelectionPopover', () => {
     expect(screen.queryByText('Test content')).not.toBeInTheDocument()
   })
 
-  it('handles mouse enter to expand selection', async () => {
+  it('does not alter selection or call onSelect when mouse enters the popover', async () => {
     const onSelect = jest.fn()
 
     render(
@@ -207,8 +207,8 @@ describe('SelectionPopover', () => {
       })
     }
 
-    // Should expand selection on mouse enter
-    expect(onSelect).toBeDefined()
+    // Hovering over the popup must NOT call onSelect or expand the selection
+    expect(onSelect).not.toHaveBeenCalled()
   })
 
   it('computes popover position correctly for desktop', () => {
--- a/quotevote-frontend/src/__tests__/components/VotingComponents/SelectionPopover.test.tsx
+++ b/quotevote-frontend/src/__tests__/components/VotingComponents/SelectionPopover.test.tsx
@@ -86,31 +86,31 @@ describe('SelectionPopover', () => {
   })
 
   it('applies custom topOffset', () => {
-    const { container } = render(
+    render(
       <SelectionPopover {...defaultProps} showPopover={true} topOffset={50} />,
     )
-    const popover = container.querySelector('#selectionPopover')
+    const popover = document.querySelector('#selectionPopover')
     expect(popover).toBeInTheDocument()
   })
 
   it('renders above the post sticky action bar (z-index > 10) by default', () => {
-    const { container } = render(
+    render(
       <SelectionPopover {...defaultProps} showPopover={true} />,
     )
-    const popover = container.querySelector('#selectionPopover') as HTMLElement
+    const popover = document.querySelector('#selectionPopover') as HTMLElement
     expect(Number(popover.style.zIndex)).toBeGreaterThan(10)
   })
 
   it('applies custom styles', () => {
     const customStyle = { zIndex: 999 }
-    const { container } = render(
+    render(
       <SelectionPopover
         {...defaultProps}
         showPopover={true}
         style={customStyle}
       />,
     )
-    const popover = container.querySelector('#selectionPopover') as HTMLElement
+    const popover = document.querySelector('#selectionPopover') as HTMLElement
     expect(popover).toHaveStyle({ zIndex: '999' })
   })
 
--- a/quotevote-frontend/src/components/VotingComponents/SelectionPopover.tsx
+++ b/quotevote-frontend/src/components/VotingComponents/SelectionPopover.tsx
@@ -116,21 +116,6 @@ export default function SelectionPopover({
     }, 100)
   }, [selectionChange])
 
-  const handleMouseEnter = useCallback(() => {
-    const selection = window.getSelection()
-    if (selection && selection.rangeCount > 0) {
-      const range = selection.getRangeAt(0)
-      // Expand to word boundary - use alternative method for Range
-      try {
-        range.setStart(range.startContainer, Math.max(0, range.startOffset - 1))
-        range.setEnd(range.endContainer, range.endOffset + 1)
-      } catch {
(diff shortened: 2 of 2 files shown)


motirebuma (approved):
### PR #401 Review — Viewport Clamping for Highlight Popup

Nice fix! The root cause was clear and the solution is clean. Here are a few quick thoughts:

**The good stuff**
- Portal via `createPortal` is exactly the right move — escapes all those pesky `overflow: hidden` ancestors.
- The new single-algorithm positioning is so much cleaner than the old 3-branch desktop/tablet/mobile split.
- Removing `handleMouseEnter` was the right call — mutating a live `Range` on hover was always going to cause trouble.
- SSR guard with `mounted` state is necessary and correct for `document` access.
- Tests updated correctly — `document.querySelector` instead of `container.querySelector` makes sense for portals.

**Small things to consider**
- The double `requestAnimationFrame` trick works, but a quick comment explaining *why two frames* would help the next person reading this.
- The `eslint-disable` comment on `setMounted` inside `useEffect` looks unnecessary — `setState` in effects is totally normal.

---

**Before merging**
- ⚠️ This PR has a merge conflict and needs a rebase onto `main` before it can land.

**Verdict: ✅ Approve (after rebase)** — solid work, good test coverage, no regressions.

motirebuma:
@OtavioXimarelli It was a simple merge conflict, and I resolved it for you. This one is ready to merge.

Thank you!.

@OtavioXimarelli @flyblackbox

Issue #378, opened 2026-06-15
RC1-026 Fix chat search by username or keyword


flyblackbox:
# RC1-026 Fix chat search by username or keyword

## Summary
Discussion/chat search should allow users to search by username or keyword.

## Area
Chat / Search

## Current Behavior
Chat search does not work as expected.

## Expected Behavior
Search returns relevant chat/discussion results for usernames and keywords.

## Tasks
- [ ] Audit chat search input and handler.
- [ ] Add support for username and keyword queries.
- [ ] Ensure query debouncing/loading states work.
- [ ] Render empty and error states clearly.

## Acceptance Criteria
- Search returns relevant chat/discussion results for usernames and keywords.
- Behavior is consistent across relevant desktop and mobile views.
- No regressions are introduced in adjacent workflows.

## Test Instructions
1. Open chat/discussion search.
2. Search by username and confirm matching results.
3. Search by keyword and confirm matching discussions/messages.
4. Confirm empty state for no results.

## Labels
`bug` `rc1` `chat` `search` `frontend` `backend`

anand-kashyap:
@flyblackbox may i pick this bug if no one else if working on it?

flyblackbox:
Yes certainly, thank you @anand-kashyap I will assign to you now.

anand-kashyap:
Hey @flyblackbox, I have my MongoDB instance running and connected, but the collections are currently empty as the README doesn't cover the setup process. Could you provide instructions on how to seed dummy data and test the user creation flow?

flyblackbox:
@motirebuma can you please provide some guidance here?  Maybe we need to update the readme

motirebuma:
@anand-kashyap  Please create two accounts in your local environment, make posts from both accounts, and test the search functionality. Then try to resolve the issue.

We don't have dummy data because things are changing over time, and maintaining mock data often creates additional confusion. That's why we've chosen not to use it for now.

Thank you!

@flyblackbox @anand-kashyap

anand-kashyap:
@motirebuma @flyblackbox it doesnt really work straight away, I had to fix a bunch of things just to get login to work. 
Please check this PR - https://github.com/anand-kashyap/quotevote-next/pull/1
and errors even in loggedin dashboard
maybe I am doing something wrong but this doesnt work out of the box

motirebuma:
@anand-kashyap I think you're using the wrong backend.

[...]

PR #402, opened 2026-06-27
fix: ChatSearchInput auto clearing on non buddy mode [RC1-026]

anand-kashyap:
RC1-026: Fix chat search input clearing during normal search**

**Branch:** `fix/chat-search-clear-guard` → `main`  
**Closes:** #378  
**Labels:** `bug` `rc1` `chat` `search` `frontend`

---

### Summary
Chat search input was clearing in normal search mode when anything is typed, which broke the ability to keep username/keyword queries in place. This fix makes the clear behavior happen only when leaving buddy-add mode.

### Problem
The search input was being cleared on initial render / non-buddy search mode instead of only when `addBuddyMode` actually changed. That made chat search behave inconsistently and interfered with searching by username or keyword.

### Fix
- Added `previousAddBuddyMode` to track the prior mode value.
- Only clear the search input when `addBuddyMode` transitions from `true` to `false`.
- Updated the comment to match the transition-based behavior.
- Added regression coverage for both normal search mode and exiting add buddy mode.

### Acceptance Criteria
- Search input remains stable in normal chat search mode.
- Search clears only when leaving add buddy mode.
- No regressions in adjacent chat workflows.

### Test Results
`pnpm test -- src/__tests__/components/Chat/ChatSearchInput.test.tsx --runInBand
```
src/__tests__/components/Chat/ChatSearchInput.test.tsx
  ChatSearchInput
    ✓ does not clear search immediately in normal search mode (26 ms)
    ✓ clears search when exiting add buddy mode (13 ms)

Test Suites: 1 passed, 1 total
Tests:       2 passed, 2 total
```

### DIFF
--- a/quotevote-frontend/src/__tests__/components/Chat/ChatSearchInput.test.tsx
+++ b/quotevote-frontend/src/__tests__/components/Chat/ChatSearchInput.test.tsx
@@ -1,0 +1,53 @@
+import { fireEvent, render, screen, waitFor } from "@/__tests__/utils/test-utils";
+import ChatSearchInput from "@/components/Chat/ChatSearchInput";
+
+describe("ChatSearchInput", () => {
+  it("does not clear search immediately in normal search mode", () => {
+    const setSearch = jest.fn();
+    const onAddBuddyModeChange = jest.fn();
+
+    render(
+      <ChatSearchInput
+        setSearch={setSearch}
+        addBuddyMode={false}
+        onAddBuddyModeChange={onAddBuddyModeChange}
+      />
+    );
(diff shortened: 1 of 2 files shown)


motirebuma (approved):
@anand-kashyap Thank for this amazing work. clean fix. Tracking the previous `addBuddyMode` value via a ref and only clearing on the `true -> false` transition is exactly the right pattern for this. Small diff, focused scope, two regression tests that cover both scenarios. Nothing to nitpick here.

Thankyou!

@anand-kashyap  @flyblackbox

PR #410, opened 2026-07-06
feat: support search by @username and #hashtag with resolvers and SVG avatar support

AegisX-dev:
## Summary
Adds post search filters for @username and #hashtag, implements missing GraphQL query resolvers, and resolves linter warnings and SVG avatar image configurations.
  
## Changes
- search: supported searching posts using @username (user ID lookup) and #hashtag (case-insensitive regex) tokens
- resolvers: implemented searchUser, group, messageRooms, getBuddyList, getRoster, and latestQuotes queries
- code quality: resolved unused imports and any types in backend resolvers
- frontend config: configured api.dicebear.com under remotePatterns and enabled dangerouslyAllowSVG in next.config.ts
- environment: enabled package builds for esbuild, prisma, and unrs-resolver in pnpm-workspace.yaml
  
## Impact
Allows users to search social content by creator username and hashtag. Prevents Next.js image loading crashes and console warnings when rendering SVG avatars.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 quotevote-backend/compose.yml
 quotevote-backend/dockerfile
+session_state.md
--- a/quotevote-backend/.gitignore
+++ b/quotevote-backend/.gitignore
@@ -29,6 +29,7 @@ npm-debug.log*
 yarn-debug.log*
 yarn-error.log*
 .pnpm-debug.log*
+*.log
 
 # env files (can opt-in for committing if needed)
 .env*
--- a/quotevote-backend/__tests__/unit/parseSearchQuery.test.ts
+++ b/quotevote-backend/__tests__/unit/parseSearchQuery.test.ts
@@ -1,0 +1,189 @@
+import { parseSearchQuery } from '~/data/utils/parseSearchQuery';
+import type { ParsedSearchQuery } from '~/types/search';
+
+describe('parseSearchQuery', () => {
+  // ── Happy path ──────────────────────────────────────────────────────────
+
+  describe('plain text queries (no tokens)', () => {
+    it('returns the full string as textQuery when no tokens are present', () => {
+      const result: ParsedSearchQuery = parseSearchQuery('hello world');
+
+      expect(result.usernames).toEqual([]);
+      expect(result.hashtags).toEqual([]);
+      expect(result.textQuery).toBe('hello world');
+      expect(result.tokens).toEqual([]);
+    });
+
+    it('trims leading and trailing whitespace from textQuery', () => {
+      const result = parseSearchQuery('  spaced out  ');
+
+      expect(result.textQuery).toBe('spaced out');
+      expect(result.usernames).toEqual([]);
+      expect(result.hashtags).toEqual([]);
+    });
+  });
+
+  // ── @username extraction ────────────────────────────────────────────────
+
+  describe('@username extraction', () => {
+    it('extracts a single @username', () => {
+      const result = parseSearchQuery('@johndoe');
+
+      expect(result.usernames).toEqual(['johndoe']);
+      expect(result.textQuery).toBe('');
+      expect(result.tokens).toEqual([{ type: 'username', value: 'johndoe' }]);
+    });
+
+    it('extracts @username and preserves remaining text', () => {
+      const result = parseSearchQuery('@alice some text here');
+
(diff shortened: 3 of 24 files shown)


motirebuma (changes requested):
Hello @AegisX-dev , This is a big PR and search parsing, six new resolvers, seed script, SVG config, tests. The `parseSearchQuery` utility is well-tested (good edge case coverage for dedup, standalone symbols, embedded @), and the `postsResolver` unit tests cover the important paths.

A few things worth addressing:

- **Regex injection in `searchUser`**: `new RegExp(queryName, 'i')` is called with raw user input. A search like `@.*` would match every user. You should escape special regex characters before constructing the pattern, same way you do it in the hashtag filter (`tag.replace(/[.*+?^${}()|[\]\\]/g, '\\$&')`).

- **N+1 queries in `rosterResolver`**: Both `getBuddyList` and `getRoster` do individual `User.findById` and `Presence.findOne` calls inside a `for` loop. With a large buddy list this will be slow. Consider batching with `User.find({ _id: { $in: buddyIds } })` instead.

- **`dangerouslyAllowSVG: true`** in next.config -- this is fine for trusted sources like dicebear, but worth a comment noting why it's safe (the `remotePatterns` restrict it to `api.dicebear.com` only).

- **CORS change**: You switched from open CORS to `origin: process.env.CLIENT_URL || 'http://localhost:3000'` with `credentials: true`. This is good security-wise, but make sure this doesn't break other environments (staging, production) that may use a different origin. The env var fallback covers it, just confirm `CLIENT_URL` is set in all deploy configs.

- **Seed script**: `seed-data.ts` calls `User.deleteMany({})` and `Post.deleteMany({})` -- this will wipe everything if someone runs it against a real database. Add a safety check (e.g., only run against localhost or a specific DB name) or at least a prominent warning in the script.

- **Scope**: This PR bundles search parsing, six resolvers, SVG config, CORS changes, workspace config, and a seed script. Consider splitting future PRs this large -- it makes review faster and reduces merge risk.

The core search logic and resolver implementations look correct. Good work on the test coverage.

Verdict: Will be approved and merged after fixing the regex injection in `searchUser` and adding a safety guard to the seed script.

Thank you!!!

@flyblackbox @AegisX-dev

AegisX-dev:
Hey @motirebuma,
  
Thank you for the detailed and constructive review! I have addressed all the feedback and pushed the updates:
  
1. Regex Injection in  searchUser : Fixed by escaping special regex characters in the search query before compiling it. I've also added unit tests validating this behavior.             
2. N+1 queries in  rosterResolver : Optimized both  getBuddyList  and  getRoster  to batch database queries utilizing  $in  query structures (dropping from 2N queries to 2). Added unit tests to cover both of these resolver flows.
3.  dangerouslyAllowSVG  in  next.config.ts : Added a comment documenting that this is safe since remote patterns restrict SVG rendering to trusted domains ( api.dicebear.com ).        
4. CORS change: Confirmed that  CLIENT_URL  is set in all deployment configurations (staging, production).
5. Seed script: Added a safety check to  seed-data.ts  that blocks execution and throws an error if the database host is anything other than  localhost  or  127.0.0.1 .
6. PR Scope: Understood. I will split future changes into smaller, more granular pull requests to make reviews faster and reduce merge risks.
  
All 761 backend tests and 1,847 frontend tests are green. Ready for your re-review and merge!***

motirebuma (approved):
Nice work @AegisX-dev. You addressed all the previous feedback cleanly:

- **Regex injection** in `searchUser` -- fixed. Special characters are now escaped before constructing the RegExp. The test that verifies `@.*` gets escaped to `@\\.\\*` is a nice touch.
- **Seed script safety guard** -- fixed. The script now refuses to run against non-local databases, which prevents accidental data wipes.
- **N+1 queries** in `rosterResolver` -- fixed. Both `getBuddyList` and `getRoster` now batch with `User.find({ _id: { $in: [...] } })` instead of individual `findById` calls in a loop.
- **`dangerouslyAllowSVG` comment** -- added, explaining it's safe because `remotePatterns` restricts SVG sources to trusted domains only.
- New unit tests for `userResolver` and `rosterResolver` -- good coverage.

CI is all green (Backend, Frontend, E2E all passing).

Verdict: approved ready to merge.

Thank you!!

@flyblackbox @AegisX-dev

Issue #394, opened 2026-06-22
5.1.5 E2E-MOD-006 Report Content Workflow


flyblackbox:
# 5.1.5 E2E-MOD-006 Report Content Workflow

## Summary

Add Playwright end-to-end coverage for the content reporting workflow.

This test should confirm that a logged-in user can report content that may violate platform rules, choose a report reason, submit the report, and receive a clear confirmation. The test should also confirm that duplicate reports are handled gracefully.

## User Story

As a user, I want to report content that violates rules.

## Test Metadata

| Field | Value |
|---|---|
| Test ID | `E2E-MOD-006` |
| Feature Area | Moderation |
| Scenario | Report Content Workflow |
| Priority | `P1` |
| Actor | Registered User |
| Auth State | Logged in |
| Required Account | `authorUser` |
| Required Data | Public post or comment |
| Viewports | Desktop, Mobile |
| Suggested Spec File | `moderation.spec.ts` |
| Source | Quote Vote User Stories |

## Test Steps

1. Log in as `authorUser`.
2. Navigate to a public post or comment that can be reported.
3. Open the report action.
4. Select a report reason.
5. Submit the report.
6. Confirm that a report confirmation appears.
7. Attempt to report the same content again.
8. Confirm that duplicate reporting is handled gracefully.

## Expected Result

The user can report content successfully.

The application should:

- Open the report flow from reportable content
- Allow the user to select a report reason
- Submit the report without errors
- Show a clear confirmation message
- Prevent duplicate reports or display a clear message if the content was already reported
- Preserve the user's current context after submission

## Required Assertions

- Reportable content loads successfully
- Report action is visible and accessible
- Report dialog or menu opens
- Report reason can be selected
- Submit button becomes available when required fields are valid
- Report submission completes successfully
- Confirmation message appears
- Duplicate report attempt does not create confusing or broken behavior
- No runtime error or generic error toast appears

## Suggested Selectors

Use stable `data-testid` selectors where possible:

- `[data-testid="post-card"]`
- `[data-testid="comment-card"]`
- `[data-testid="report-button"]`
- `[data-testid="report-dialog"]`
- `[data-testid="report-reason-option"]`
- `[data-testid="report-submit-button"]`
- `[data-testid="report-confirmation"]`
- `[data-testid="report-duplicate-message"]`

If these selectors do not exist yet, add them as part of the test implementation.

## Out of Scope

This issue does not need to test:

- Moderator review queue behavior
[...]

PR #424, opened 2026-07-13
fix(backend): duplicate featuredSlot index + seed script wrong DB

CynthiaWahome:
# Pull Request Template

## Description

Two small, unrelated backend hotfixes found while doing local setup for #394/#395, bundled here since both are one-line-scale config/typo fixes with no shared code:

1. **`Post.ts`** declared the same unique+sparse index on `featuredSlot` twice — once via field-level `unique`/`sparse` options, once via an explicit `PostSchema.index(...)` call. Mongoose warns about this on every server boot. Removed the redundant field-level flags; the explicit `.index()` call (already grouped with the other `featuredSlot` indexes) remains the single source of truth, so behavior is unchanged.
2. **`scripts/seed-data.ts`** never called `dotenv.config()`, so `process.env.MONGO_URI` was always empty when run directly, silently falling back to its hardcoded default database name instead of whatever `.env` specifies. If that default differs from the database the running server actually connects to, seeded users land in the wrong database — the app then reports "invalid credentials" for logins that should work, with no indication the seed data even landed somewhere else. Added `dotenv.config()` so it reads the same `.env` the server does.

(A third fix found in the same session — a completely missing `user(username)` resolver — was split out into #425 since it's a larger, distinct change implementing a previously-unbuilt feature rather than a config fix.)

## Related Issue (Link to issue ticket)

None — found doing local setup for #394/#395, filed as standalone hotfixes rather than bundled into that unrelated work.

## Motivation and Context

**(1)** Not a functional bug, but trains contributors to skim past console warnings — a habit that causes real ones to get missed later.

**(2)** This one cost real debugging time: seeded a local dev DB, tried logging in with the seeded credentials, got a generic "invalid credentials" error that gave no hint the actual problem was two different databases in play. Anyone else running this script for local setup would hit the same silent trap.

**Before (`Post.ts`, every `pnpm dev` startup):**
```
(node:91073) [MONGOOSE] Warning: Duplicate schema index on {"featuredSlot":1} found. This is often due to declaring an index using both "index: true" and "schema.index()". Please remove the duplicate index definition.
```
**After:** warning no longer prints; index behavior (uniqueness + sparse) unchanged.

**Before (`seed-data.ts`):** running `npx tsx scripts/seed-data.ts` with `.env` set to `MONGO_URI=mongodb://localhost:27017/quotevote-dev` actually seeded `mongodb://localhost:27017/quotevote` (the hardcoded fallback, no `.env` loaded) — then logging in against the real server (connected to `quotevote-dev`) failed:
```
curl -X POST http://localhost:4000/auth/login -d '{"username":"alice@quote.vote","password":"securepassword123"}'
{"message":"Invalid username or password."}
```
**After:** script loads `.env`, seeds the correct database, login succeeds:
```
curl -X POST http://localhost:4000/auth/login -d '{"username":"alice@quote.vote","password":"securepassword123"}'
{"accessToken":"eyJhbGciOiJIUzI1NiIs...","user":{"email":"alice@quote.vote",...}}
```

## How Has This Been Tested?

- Reproduced both issues locally against a Dockerized MongoDB before fixing.
- After fix (1): confirmed only one `{featuredSlot:1}` index declaration with matching options remains in the schema.
- After fix (2): re-ran the seed script, confirmed via direct DB query that users land in the database specified by `.env`, then confirmed login succeeds end-to-end via `curl` against the running server (see output above).

## Screenshots (if appropriate - Postman, etc)

N/A — terminal output included above.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed (backend CI green; unrelated pre-existing frontend test failure noted below, not caused by this change).

## Note on CI

`Frontend CI` fails on this PR due to a pre-existing, unrelated Jest test (`ProfileHeaderMessage.test.tsx` expecting a `username` field no longer present on the message object). Proof it's unrelated to this PR: this PR only touches two backend files (`quotevote-backend/app/data/models/Post.ts`, `quotevote-backend/scripts/seed-data.ts`); diffing both the failing test file and the component it tests against `upstream/main` shows **zero difference**:

```
git diff upstream/main <this-branch> -- quotevote-frontend/src/__tests__/components/Profile/ProfileHeaderMessage.test.tsx
git diff upstream/main <this-branch> -- quotevote-frontend/src/components/Profile/ProfileHeader.tsx
# both empty — byte-identical to main
```

This failure exists on `main` right now, independent of this PR. `Vercel` shows "Authorization required to deploy," which is a fork-PR permissions gate, not a build failure. `Backend CI` (the only check covering this PR's actual changes) passes.

### DIFF
--- a/quotevote-backend/app/data/models/Post.ts
+++ b/quotevote-backend/app/data/models/Post.ts
@@ -26,8 +26,6 @@ const PostSchema = new Schema<PostDocument, PostModel>(
       type: Number,
       min: 1,
       max: 12,
-      unique: true,
-      sparse: true,
     },
     messageRoomId: { type: String },
     urlId: { type: String },


motirebuma (changes requested):
Hi @CynthiaWahome, the duplicate index fix is clean and worth merging. Removing the field-level `unique`/`sparse` flags while keeping the explicit `.index()` call eliminates the Mongoose warning on every startup. Good catch.

The seed script dotenv fix itself works, but the underlying seed script has some issues worth addressing:

- **Plaintext passwords** — Users are created with `password: 'securepassword123'` without hashing. Unless the User model's `pre('save')` hook handles this, logins against these accounts won't work.
- **Nonexistent group** — Posts reference a `dummyGroupId` (random ObjectId) that doesn't correspond to any actual group, so seeded posts won't appear correctly in group views.
- **Destructive wipe** — `User.deleteMany({})` and `Post.deleteMany({})` nuke all existing data. Combined with the dotenv fix now targeting the correct DB, **this actually becomes more dangerous. before, it was wiping the wrong (empty) database silently. Now it'll wipe the one you're actually using.**
The dotenv fix makes the script hit the right database, but the seed data it creates isn't usable for local development anyway. Consider either improving the seed script to create realistic data (hashed passwords, real groups) or documenting these limitations clearly.

Verdict: the index fix is ready to merge. The seed script dotenv fix is technically correct but the script itself needs improvement to be actually useful.

Thank you!!

@flyblackbox @CynthiaWahome

flyblackbox:
Hello @CynthiaWahome just checking back on this one, do you want us to merge?  Or do you want to implement any of the suggestions from @motirebuma?   Let us know what you think of the feedback, and thank you for your excellent contribution!

CynthiaWahome:
Hi @motirebuma 
Thanks for the feedback

Hi @flyblackbox 
I can work on the seed script on another DB, so we can have separation of concerns, if that okay with you.

flyblackbox:
I'm not sure @CynthiaWahome but if you and @motirebuma agree, than I am supportive.

flyblackbox:
@CynthiaWahome @motirebuma the discussion suggests two changes with different risk levels are currently coupled in this PR:

- the duplicate `featuredSlot` index cleanup appears ready
- the dotenv change makes the destructive seed script target the intended database, which increases the importance of the unresolved password, group-reference, and data-wipe concerns

A low-risk path may be to keep the index fix here, remove the seed-script dotenv change from this PR, and track seed-script hardening separately. Could you confirm whether you agree with that split?

If so, @CynthiaWahome, please rebase onto current `main`, resolve the conflicts, and narrow this branch to the index correction. If you prefer to retain the seed change, please add safeguards or documentation that make the target database and destructive behavior unmistakable before merge.

CynthiaWahome:
@flyblackbox Agreed on the split. I've rebased onto main and narrowed this branch to just the featuredSlot index fix — the seed-script dotenv change is dropped from this PR, I'll track hardening it (password hashing, real group refs, safer wipe behavior) separately. Ready for another look.

motirebuma (approved):
Hey @CynthiaWahome, great job addressing the feedback! You removed the seed script entirely and kept only the index fix. This is exactly what was requested.

The diff is now just two lines removing `unique: true` and `sparse: true` from the `featuredSlot` field in `Post.ts`. The explicit `.index()` call elsewhere handles the indexing correctly. Clean and minimal.

No issues.

Verdict: **approve**

LGTM

@flyblackbox @CynthiaWahome

Issue #395, opened 2026-06-22
5.1.6 E2E-MOD-007 Report User as Bot


flyblackbox:
# 5.1.6 E2E-MOD-007 Report User as Bot

## Summary

Add Playwright end-to-end coverage for reporting a user account as a suspected bot.

This test should confirm that a logged-in user can open another user’s profile or user menu, choose the report option, select “Bot / automated account” as the reason, submit the report, and receive a clear confirmation.

This is a P1 test because bot-reporting supports platform integrity, but the flow should remain careful and non-punitive. A report should create a moderation signal, not automatically restrict or label the reported user.

## User Story

As a user, I want to report an account that appears to be a bot so moderators can review suspicious activity.

## Test Metadata

| Field | Value |
|---|---|
| Test ID | `E2E-MOD-007` |
| Feature Area | Moderation |
| Scenario | Report User as Bot |
| Priority | `P1` |
| Actor | Registered User |
| Auth State | Logged in |
| Required Account | `authorUser` |
| Required Data | Reportable user profile |
| Viewports | Desktop, Mobile |
| Suggested Spec File | `moderation.spec.ts` |
| Source | Generated |

## Test Steps

1. Log in as `authorUser`.
2. Navigate to another user’s profile.
3. Open the user action menu or report control.
4. Select the report user option.
5. Choose `Bot / automated account` as the report reason.
6. Submit the report.
7. Confirm that a report confirmation appears.
8. Attempt to report the same user again.
9. Confirm that duplicate bot reports are handled gracefully.
10. Repeat the test in desktop and mobile Playwright projects.

## Expected Result

The user can report another account as a suspected bot.

The application should:

- Open the user report flow from a profile or user action menu
- Allow the reporter to select `Bot / automated account`
- Submit the report without errors
- Show a clear confirmation message
- Preserve the reporter’s current context
- Prevent duplicate reports or display a clear already-reported message
- Avoid publicly labeling the reported user as a bot based only on the report

## Required Assertions

- Reportable user profile loads successfully
- User report action is visible and accessible
- Report dialog or menu opens
- `Bot / automated account` reason is available
- Bot reason can be selected
- Submit button becomes available when required fields are valid
- Report submission completes successfully
- Confirmation message appears
- Duplicate report attempt does not create confusing or broken behavior
- Reported user profile remains viewable unless moderation state already requires otherwise
- No runtime error or generic error toast appears

## Suggested Selectors

Use stable `data-testid` selectors where possible:

[...]

PR #425, opened 2026-07-13
fix(backend): implement missing user(username) query resolver

CynthiaWahome:
# Pull Request Template

## Description

The GraphQL schema declares `user(username: String!): User` (`app/server.ts`), and the frontend's `ProfileController.tsx` queries it on every single profile page view — but no resolver implementation existed anywhere in the codebase for this field. Wired it up using the already-written (but never-called) `User.findByUsername()` static on the model.

## Related Issue (Link to issue ticket)

None directly — found while doing local setup/manual QA for #395, which requires viewing another user's profile to trigger the report-user dialog. Filed standalone since it's a distinct fix (implementing an unbuilt feature) from smaller config-typo fixes.

## Motivation and Context

Without a resolver, Apollo Server's default behavior returns `null` for the field instead of erroring — so the frontend got a clean HTTP 200 with `{"data":{"user":null}}`, and silently rendered its "Invalid user" fallback UI. No error anywhere, in any log, made this look like a client bug rather than a missing backend implementation. This was blocking any manual verification of profile-dependent flows, including #395.

**Before:**

Invalid user error

```
GET /dashboard/profile/charlie 200 in 30ms
```
(200 status, but page renders "Invalid user" — response body was `{"data":{"user":null}}`, no GraphQL errors array)

image

---

**After:**

Charlie's real profile rendering

```
curl -X POST http://localhost:4000/graphql -d '{"query":"query { user(username: \"charlie\") { _id username name } }"}'
{"data":{"user":{"_id":"6a545c664d951ee86a637b1e","username":"charlie","name":"Charlie Chaplin"}}}
```
Screenshot 2026-07-13 at 07 08 18

## How Has This Been Tested?

- Reproduced locally: viewing any profile other than the logged-in user's own rendered "Invalid user" despite a 200 response.
- After fix: confirmed via direct GraphQL query (`curl`, shown above) and in the browser — profile page renders real user data, follower/following counts, and their posts.

## Screenshots (if appropriate - Postman, etc)

See before/after above.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.

### DIFF
--- a/quotevote-backend/__tests__/unit/resolvers/userResolver.test.ts
+++ b/quotevote-backend/__tests__/unit/resolvers/userResolver.test.ts
@@ -21,15 +21,17 @@ describe('userResolver', () => {
 
     it('escapes special regex characters to prevent injection', async () => {
       (User.find as jest.Mock).mockReturnValue({
-        limit: jest.fn().mockReturnValue({
-          lean: jest.fn().mockResolvedValue([
-            {
-              _id: new mongoose.Types.ObjectId('60d5ec49ad414d7a8d5464a0'),
-              name: 'Alice Cooper',
-              username: 'alice',
-              accountStatus: 'active',
-            },
-          ]),
+        select: jest.fn().mockReturnValue({
+          limit: jest.fn().mockReturnValue({
+            lean: jest.fn().mockResolvedValue([
+              {
+                _id: new mongoose.Types.ObjectId('60d5ec49ad414d7a8d5464a0'),
+                name: 'Alice Cooper',
+                username: 'alice',
+                accountStatus: 'active',
+              },
+            ]),
+          }),
         }),
       });
 
--- a/quotevote-backend/app/data/resolvers/userResolver.ts
+++ b/quotevote-backend/app/data/resolvers/userResolver.ts
@@ -3,6 +3,18 @@ import type * as Common from '~/types/common';
 
 export const userResolver = {
   Query: {
+    user: async (
+      _parent: unknown,
+      args: { username: string }
+    ): Promise<Common.User | null> => {
+      const user = await User.findByUsername(args.username).lean();
+      if (!user) return null;
+
+      return {
+        ...user,
+        _id: user._id.toString(),
+      } as unknown as Common.User;
+    },
     searchUser: async (
       _parent: unknown,
       args: { queryName: string }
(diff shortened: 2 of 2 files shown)


CynthiaWahome:
Addressed in 7bcc3d0:
- Restricted the resolver to a `.select()` of public-profile fields only (`_id name username avatar contributorBadge upvotes downvotes _followingId _followersId reputation`) — `email` is no longer returned even if explicitly requested, since this is an unauthenticated query per `requireAuth.ts`'s `PUBLIC_QUERIES` list.
- Added `accountStatus: 'active'` filter to match `searchUser`'s existing behavior.

Verified: `user(username:"bob"){ email }` now returns `email: null` instead of the real address; other fields unaffected.

Holding off on unit tests for this PR to stay focused on unblocking #394/#395 today — will follow up separately if that's the right call, open to adding them if preferred before merge.

CynthiaWahome:
Addressed the second round of findings in 86b93b3:
- Trimmed the `username` arg before querying.
- Applied the same public-field `.select()` to `searchUser` — you were right that fixing only `user()` left the identical email-harvesting vector open there.

Unit test coverage for `Query.user` is still outstanding — deferring that to keep moving on #394/#395 today, will follow up.

motirebuma (approved):
Nice work @CynthiaWahome. This was a real blocker, the schema declared `user(username: String!): User` and the frontend queries it on every profile page, but no resolver existed. Apollo silently returned `null` with a 200 status, making it look like a client bug. Good detective work finding this.

The implementation is solid:

- **Field selection** — Only exposing public-profile fields (`_id`, `name`, `username`, `avatar`, `contributorBadge`, `upvotes`, `downvotes`, `_followingId`, `_followersId`, `reputation`) via `.select()` is good security practice. Prevents leaking emails or other sensitive data through this unauthenticated query.
- **Active-account filter** — Filtering by `accountStatus: 'active'` is consistent with `searchUser`.
- **Reputation `_id` handling** — Clean workaround for the `UserReputationType` requiring `_id` on the embedded 1:1 subdocument.
- **Consistency** — Adding `.select()` to the existing `searchUser` resolver too is a nice touch.
- **Test update** — Updated mock chain to include `.select()` before `.limit()`.

Thank you!!

@flyblackbox @CynthiaWahome

Issue #422, opened 2026-07-11
5.1.11 E2E-AUTH-001 Signup / Account Creation


flyblackbox:
# 5.1.11 E2E-AUTH-001 Signup / Account Creation

## Summary

Add Playwright end-to-end coverage for creating a new Quote.Vote account.

This test should confirm that a logged-out visitor can open the signup flow, enter a unique username, email address, and valid password, submit the form, and reach the authenticated experience.

This is a P0 test because account creation is a core participation path. A failure in this flow prevents new users from joining Quote.Vote.

## User Story

As a visitor, I want to create an account so I can participate as an accountable user.

## Test Metadata

| Field | Value |
|---|---|
| Test ID | `E2E-AUTH-001` |
| Feature Area | Authentication |
| Scenario | Signup / Account Creation |
| Priority | `P0` |
| Actor | Visitor |
| Auth State | Logged out |
| Required Data | Unused email and username |
| Viewports | Desktop, Mobile |
| Suggested Spec File | `auth.spec.ts` |
| Source | Generated |

## Test Steps

1. Open Quote.Vote while logged out.
2. Navigate to the signup flow.
3. Enter a unique username.
4. Enter a unique email address.
5. Enter a valid password.
6. Submit the signup form.
7. Confirm the account is created.
8. Confirm the new user becomes authenticated.
9. Confirm the application routes to the authenticated experience.
10. Repeat the test in desktop and mobile Playwright projects.

## Expected Result

The visitor creates an account successfully.

The application should:

- Accept an unused username
- Accept an unused email address
- Accept a valid password
- Create the account
- Authenticate the new user
- Route the user to the authenticated experience
- Avoid duplicate account creation
- Avoid runtime errors or generic error toasts

## Required Assertions

- Signup page or dialog loads successfully
- Username input accepts text
- Email input accepts text
- Password input accepts a valid password
- Submit button becomes available when required fields are valid
- Signup completes successfully
- Authenticated navigation or user state appears
- User is redirected to the expected authenticated route
- Session persists after page reload
- No runtime error or generic error toast appears

## Suggested Selectors

Use stable `data-testid` selectors where possible:

- `[data-testid="signup-form"]`
- `[data-testid="signup-username-input"]`
- `[data-testid="signup-email-input"]`
- `[data-testid="signup-password-input"]`
- `[data-testid="signup-submit-button"]`
- `[data-testid="authenticated-navigation"]`
- `[data-testid="user-profile-menu"]`

[...]

PR #426, opened 2026-07-14
implemented E2E-AUTH-001 signup test coverage and fixed test suite issues

sandy4242:
issue link: #422 

- Added stable selectors to the signup form` (data-testid="signup-form", input fields, and submit button)` inside `PageContent.tsx`
- Added authenticated layout selectors `(data-testid="authenticated-navigation", data-testid="user-profile-menu")` to 
`layout.tsx` for both desktop and mobile viewports.
- Updated the signup form submission to try `/auth/register` (the active backend path) first, falling back to `/register `dynamically to align with backend endpoint migrations.
- Added a spec targeting the logged-out visitor signup, redirects, ToS/CoC checkbox login, landing on the dashboard page, and session persistence across page reloads.
- Refined the generated test username to `user_${Date.now().toString().slice(-10)}` to strictly satisfy the Zod maximum length limit of 20 characters.
- Fixed the `hasAuthCookie `reference error in the duplicate invite spec in `auth.spec.ts`
### `TESTs`
- Updated `setSelectedChatRoom `expectation inside `ProfileHeaderMessage.test.tsx`
 to align with actual component return values, securing a clean run across all `157 Jest test suites`.
- Fixed pre-existing code error: Restored the missing cookie variable lookup `(hasAuthCookie)` inside the "handles duplicate invite request gracefully" test block to fix a syntax reference error.

`test and test issues faced were solved using ai`

### DIFF
--- a/quotevote-frontend/e2e/auth.spec.ts
+++ b/quotevote-frontend/e2e/auth.spec.ts
@@ -216,3 +216,92 @@ test.describe("Invite Request CTA (E2E-AUTH-005)", () => {
     expect(hasAuthCookie).toBe(false);
   });
 });
+
+test.describe("Signup / Account Creation (E2E-AUTH-001)", () => {
+  // Override global storage state so we start as a logged-out visitor
+  test.use({ storageState: { cookies: [], origins: [] } });
+
+  test("creates a new account, logs in, and routes to authenticated dashboard", async ({ page }) => {
+    const errorToastLocator = page.locator('[data-sonner-toast][data-type="error"]');
+    const uniqueId = Date.now();
+    const uniqueUsername = `user_${uniqueId.toString().slice(-10)}`;
+    const uniqueEmail = `e2e-signup-${uniqueId}@example.com`;
+    const validPassword = `Password${uniqueId}!`;
+
+    console.log(`TEST PROGRESS: Generated uniqueUsername=${uniqueUsername}, uniqueEmail=${uniqueEmail}`);
+
+    // 1. Open Quote.Vote while logged out and navigate to signup
+    console.log("TEST PROGRESS: Navigating to signup page");
+    await page.goto("/auths/signup");
+
+    // 2. Assert signup form and elements load successfully
+    console.log("TEST PROGRESS: Asserting signup form elements are loaded");
+    const signupForm = page.getByTestId("signup-form");
+    const usernameInput = page.getByTestId("signup-username-input");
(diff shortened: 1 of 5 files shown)


sandy4242:
@flyblackbox hi,  pr related to #422

motirebuma (changes requested):
Hey @sandy4242 good scope here: the signup E2E test, data-testid selectors, and the ProfileHeaderMessage test fix (which resolves the Frontend CI failure that's been broken on main). A few things to clean up:

1. **Remove the `console.log` debug statements.** There are 10+ `console.log("TEST PROGRESS: ...")` calls throughout the E2E test. Playwright has built-in test reporting and tracing — use those instead of console.log for debugging. These should not be committed.

2. **Use the correct registration endpoint — don't try both.** The backend uses `/auth/register` (you can verify this in `quotevote-backend/app/server.ts` where the auth routes are mounted). The current code tries `/auth/register` first, and if it fails, retries with `/register`. This means every signup makes two network requests when the first one fails for any reason (validation error, duplicate email, etc.) — since the fallback can't distinguish "wrong endpoint" from "legitimate error." Just use `/auth/register` directly and remove the fallback block entirely.

3. **The `pageerror` listener is registered too late.** You set up `page.on("pageerror", ...)` at step 10 after the entire test has already run. Any page errors from steps 1–9 won't be caught. Move the listener to the top of the test, before any navigation.

4. **Missing newline at end of `auth.spec.ts`.**

The ProfileHeaderMessage test fix (adding the `username` field to match the StagedChatRoom type update) is correct and fixes the CI failure on main, nice catch.

Verdict: Will be approved and merge after removing the console.log statements, fixing the pageerror listener placement, and reconsidering the endpoint fallback approach.

sandy4242:
@motirebuma fixed all the mentioned blockers, please continue reviewing, thanks

motirebuma (approved):
Nice work @sandy4242. You addressed all the previous feedback:

- **Console.log statements** — removed. Clean.
- **Endpoint fallback** — removed. Now uses `/auth/register` directly instead of trying both endpoints.
- **Pageerror listener** — fixed in the new signup test (moved to the top of the test before any navigation).

The signup E2E flow is solid: account creation → login with ToS/CoC → dashboard verification → session persistence after reload. The ProfileHeaderMessage test fix is correct and resolves the CI failure on main.

CI is all green.

Thank you @sandy4242 

@flyblackbox @sandy4242

Issue #138, opened 2025-11-28
7.30 Migrate GraphQL Enums and Relationships to TypeScript


flyblackbox:
# 7.30 Migrate GraphQL Enums and Relationships to TypeScript

## Summary
Migrate GraphQL enum and relationship type definitions to TypeScript, ensuring alignment with Prisma enums, domain constants, and relationship fields across the schema.

## Tasks
- [ ] Identify all GraphQL enums and relationship-related types in `app/data/types/`
- [ ] Convert their definitions from `.js` to `.ts`:
      - Use `GraphQLEnumType` with TypeScript typing
      - Mirror enum values from shared constants or Prisma enums
- [ ] Align enum definitions with:
      - `app/data/utils/constants.ts`
      - Prisma enum types in `prisma/schema.prisma`
- [ ] Ensure relationship-related GraphQL types correctly reference:
      - User ↔ Post / Comment / Vote / Presence / Roster
      - Post ↔ Comment / Quote / Vote
      - Message ↔ MessageRoom / Reaction / TypingIndicator / Presence
      - Group ↔ User / Roster
- [ ] Update imports where enums or relationship types are used in type files and resolvers
- [ ] Run TypeScript compilation and GraphQL schema build
- [ ] Commit updated enum and relationship type files

## Expected Behavior
GraphQL enums and relationship types live in TypeScript modules, align with Prisma enums and domain constants, and support resolvers that rely on these values and links.

## Tech Notes
- Use a single source of truth for enum values to avoid drift (constants → Prisma → GraphQL).
- Keep relationship typing consistent with Prisma relation fields and shared domain types.

## Acceptance Criteria
- All GraphQL enums and relationship-related types compile as `.ts` modules
- Enum values align with constants and Prisma enums
- Relationship typing in GraphQL schema matches underlying Prisma relationships
- Schema introspection shows expected enum values and type relationships

## Test Instructions
1. Run `npx tsc --noEmit` to validate enum and relationship type definitions.
2. Introspect the GraphQL schema (via GraphiQL or tooling) to confirm enum definitions and relationships.
3. Execute GraphQL queries and mutations that rely on enums and relationships (e.g., filter by status, role, type).
4. Confirm behavior aligns with existing application expectations.

## Labels
`backend` `graphql` `typescript` `enums` `schema` `phase7`

sandy4242:
@flyblackbox this would be a great issue to work on, thank you

flyblackbox:
[...]

PR #430, opened 2026-07-19
Migrate GraphQL enums and relationships to TypeScript

sandy4242:
- Created shared constants in `constants.ts` mapping to Prisma/GraphQL enum values, and defined matching programmatic `GraphQLEnumType` instances in `enums.ts` (such as `AccountStatusEnum`, `VoteTypeEnum`, etc.).
-  Implemented database-backed resolvers linking core types `(User, Post, Comment, Vote, Message, MessageRoom, Group, Roster, Presence, TypingIndicator, Notification, Reaction, Activity, UserInvite, UserReport)`
- `TESTS WERE DONE USING AI` and `no errors found`

issue link - #138


sandy4242:
@flyblackbox

motirebuma (changes requested):
Hey @sandy4242, the structure of this PR is clean. Shared constants in `constants.ts`, programmatic `GraphQLEnumType` instances in `enums.ts`, and database-backed resolvers across all the core types. The `toEnumValues` helper is a nice touch for keeping the enum definitions DRY. Expanding `AccountStatus` in `common.ts` to include `'suspended' | 'pending'` is correct. Test coverage is updated. CI is all green.

Three things need attention:

1. **N+1 query problem.** Every relationship field does a separate `Model.findById()` or `Model.find()` call. When a client queries a list of posts, each post individually hits the DB for its creator, comments, votes, quotes, and messageRoom. At scale this will be a serious performance issue. This doesn't need to block the PR, but it should be tracked as follow-up work — a DataLoader layer (or batch query strategy) is the standard fix for this in GraphQL.

2. **`Group.rosters` resolver looks wrong.** You have:
   ```ts
   rosters: {
     type: new GraphQLList(RosterType),
     resolve: (group) => Roster.find({ userId: group.creatorId }).lean(),
   }
   ```
   This finds rosters where the *creator* is the user, not rosters belonging to the group. Should this be querying by `groupId` or by the group's member list instead?

3. **Enum values added to constants but not to `common.ts` type unions.** You updated `AccountStatus` in `common.ts`, but the constants also add new values for `RosterStatus` (`'declined'`), `MessageType` (`'SYSTEM'`), and others that don't have matching updates in the TypeScript union types. The GraphQL enums will accept these values at runtime, but TypeScript won't know about them — which defeats the purpose of the migration. Please sync the `common.ts` type unions with the new constant values.

Also a minor note: the PR description says "TESTS WERE DONE USING AI and no errors found." The tests do pass in CI, but please double-check that the test assertions are actually verifying the new enum values and resolver behavior, not just checking that the types compile.

Verdict: address items 1, 2 and 3 above, then this is ready to merge.

Thank you @sandy4242 

@flyblackbox @sandy4242

sandy4242:
@motirebuma @flyblackbox 
the test previously failed, later were fixed by me and a bit of `ai`

motirebuma (approved):
Hey @sandy4242 good work addressing the feedback. Both requested changes are fixed:

1. **Group.rosters resolver — fixed.** Now queries rosters where any group member (creator + allowed users) appears as either `userId` or `buddyId`. The `$or` query with `$in` on the combined member list is the right approach. The new test verifies this exact query shape.

2. **common.ts type unions — fixed.** `NotificationType`, `MessageType`, `RosterStatus`, `AccountStatus` are all updated to match the constant values. `InviteStatus` type added. `Message.type` properly typed from `string` to `MessageType`. Good.

The new test coverage is solid — runtime enum value validation against the schema, and a resolver integration test with proper model mocking.

One note for future work: the N+1 query pattern (every relationship field doing a separate `findById`) will need a DataLoader layer at scale. Not a blocker for this PR, but worth tracking as a follow-up issue.

Thank you @sandy4242 

@flyblackbox @sandy4242

PR #433, opened 2026-07-22
fix(frontend): featured posts navigate to post detail (RC1-004) #355

### LINKED ISSUE #355: RC1-004 Featured posts should navigate to post detail page
# RC1-004 Featured posts should navigate to post detail page

## Summary

Featured post cards now navigate to the correct post detail page.

## Resolution

Implemented and merged through #433 on July 22, 2026.

## Completed work

- [x] Reviewed the featured-post card and link implementation.
- [x] Routed cards to the correct post detail URL using the shared route helper.
- [x] Made the featured-card interaction consistent.
- [x] Added unit regression coverage.
- [x] Added Playwright coverage for desktop and mobile navigation.

## Acceptance evidence

- #433 is merged to `main`.
- The merged change includes `quotevote-frontend/e2e/featured-posts.spec.ts`.
- The merged change includes featured-post navigation cases in `src/__tests__/app/page.test.tsx`.

## Product verification

Additional exploratory browser testing may still be useful during RC1 validation, but the implementation and regression-coverage scope of this issue is complete.

niamao:
## Summary
- Make each featured post card a link to its post detail route via `toAppPostUrl`.
- Add unit and E2E regression coverage for featured-post navigation (desktop + mobile).
- Stop tracking `.env.local` (add `.env.example` + gitignore) so API overrides stay machine-local.
- Fix `GetActionReactions` variable type (`String!`) to match the hosted API and avoid 400s.
- Request activity `_id` and use stable keys in activity lists (no index-based keys).
- Ignore local `.cursor/` so IDE/agent rules stay machine-specific.

## Commits
1. `fix(frontend): navigate featured posts to detail pages (RC1-004)`
   - Featured post card → post detail navigation
   - Unit + E2E coverage
   - Untrack `.env.local`; add `.env.example` and frontend gitignore/README updates
2. `fix: align actionReactions types and stabilize activity list keys`
   - `GetActionReactions($actionId: String!)` (frontend + backend schema)
   - `GET_USER_ACTIVITY` requests entity `_id`
   - Stable React keys in `ActivityList` / `PaginatedActivityList`
3. `chore: ignore local .cursor directory`
   - Root `.gitignore` entry for `.cursor/`

## Test plan
- [ ] Open homepage logged out; confirm Featured Posts cards are links.
- [ ] Click each featured card and land on `/dashboard/post/...` for that post.
- [ ] Confirm browser back returns to the homepage with cards still visible.
- [ ] Open a post with comments/votes; confirm no `GetActionReactions` 400 in the console.
- [ ] Open a profile activity feed; confirm no React unique-key warning.
- [ ] `pnpm test -- src/__tests__/app/page.test.tsx` (featured posts navigation cases)
- [ ] `pnpm test:e2e featured-posts.spec.ts`
- [ ] Confirm `.env.local` is ignored and new clones use `cp .env.example .env.local`
- [ ] Confirm `.cursor/` is ignored and not present in the PR diff

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,6 @@ quotevote-backend/compose.yml
 quotevote-backend/dockerfile
 session_state.md
 graphify-out/
+
+# Local Cursor IDE rules / agent config (machine-specific)
+.cursor/
--- a/quotevote-backend/app/server.ts
+++ b/quotevote-backend/app/server.ts
@@ -112,8 +112,8 @@ async function startServer() {
         getBuddyList: [BuddyWithPresence!]!
         getRoster: [Roster!]!
         
-        # Action reactions
-        actionReactions(actionId: ID!): [Reaction!]!
+        # Action reactions (String! — matches hosted/legacy API; ID! breaks clients)
+        actionReactions(actionId: String!): [Reaction!]!
         
         # Admin / reports
         getBotReportedUsers(sortBy: String, limit: Int): [User!]!
--- a/quotevote-frontend/.env.example
+++ b/quotevote-frontend/.env.example
@@ -1,0 +1,14 @@
+# Quote.Vote API — local development
+# Copy to `.env.local` and adjust for your machine. Never commit `.env.local`.
+NEXT_PUBLIC_SERVER_URL=http://localhost:4000
+NEXT_PUBLIC_GRAPHQL_ENDPOINT=http://localhost:4000/graphql
+
+# Optional: point at hosted staging/production instead of local backend
+# NEXT_PUBLIC_SERVER_URL=https://api.quote.vote
+# NEXT_PUBLIC_GRAPHQL_ENDPOINT=https://api.quote.vote/graphql
+
+# WebSocket is auto-derived by getGraphqlWsServerUrl():
+#   ws://localhost:4000/graphql   (local)
+#   wss://api.quote.vote/graphql  (hosted)
+# Login is a REST endpoint (not GraphQL):
+#   POST {NEXT_PUBLIC_SERVER_URL}/auth/login  →  { username, password }  →  { token }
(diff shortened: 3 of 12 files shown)


motirebuma (approved):
hi @niamao nice work all-around, clean PR. Looks good to me overall.

My thoughts on your changes:

1. Featured post card nav (RC1-004): wrapping the card in a  when a post.url is present, and a  otherwise, is the right approach. Using toAppPostUrl is good for consistency, and I also like that you've added the necessary aria-label and focus-visible styling. Using the shared cardContent fragment to avoid duplication between the link and div cases is a good touch as well.

2. actionReactions type: changing actionId: ID! to String! in both the backend and the frontend query is the correct fix here. ID is a scalar type that can be serialized differently on different clients, but the hosted API (and presumably legacy APIs) expect a string. The change makes sense, and the note in the commit message explaining why it's String! is a good addition.

3. Stable activity list keys: using item._id along with the postId fallback, activity type, vote id, etc. is the right approach. Adding _id to the GET_USER_ACTIVITY query (and the corresponding activity type and vote id fields) ensures that we have the unique key available in all cases. I think this is sufficient, but in the rare case that those fields are somehow missing, it'll fall back to the less reliable combination of postId + activityType + voteId, which should still be better than using the array index. This pattern has been applied to both ActivityList.tsx and PaginatedActivityList.tsx.

4. .env.local -> .env.example: unstaging the .env.local file and instead adding a more informative .env.example is the right thing to do. Looks like the .gitignore patterns are also correct (ignoring .env.local, but not .env.example and .env.e2e.example). README has been updated as well.

5. .cursor/ in .gitignore: no problem, that's a sensible addition.

Testing: I've reviewed both the unit and E2E tests, and they both look good to me. The unit tests verify that the card is rendered as an  tag with the expected href, and confirm that data-testid selectors are present. The E2E test mocks GraphGL and fires off a click event, verifying that the card navigation works and then the back button works as well. The note about clearing storageState to emulate a logged-out session is a good touch.

My general impression: looks good to me, approved for merging.

Thanks @niamao

@flyblackbox @niamao

PR #442, opened 2026-07-31
fix: resolve emoji reactions query and mutations for quote/vote actions

AegisX-dev:
## Summary
Fixes **RC1-016**: Resolves emoji reaction rendering, query handling, and mutation persistence for quote and vote activity records.
  
### Backend (`quotevote-backend`)
- **Resolvers**: Created reactionResolver.ts to handle:
- `actionReactions` query
- `addActionReaction` mutation
- `updateActionReaction` mutation
- **Schema & Server**: Added mutation definitions (`addActionReaction`, `updateActionReaction`) to server.ts and registered `reactionResolver`.
- **Unit Tests**: Created reactionResolver.test.ts covering query and mutation execution paths under auth guard conditions.
  
### Frontend (`quotevote-frontend`)
- **State Guard**: Added optional chaining in CommentReactions.tsx when reading user ID to prevent crashes in unauthenticated or empty store states.
- **Event Propagation**: Stopped click event bubbling on reaction component container to prevent unexpected card selection expansion when toggling reactions.
  
## Verification & Testing
- Ran backend test suite: **99 passed / 99 total**
- Ran frontend test suite: **159 passed / 159 total**
- Verified reactions render, update, and persist correctly across action types.

### DIFF
--- a/quotevote-backend/__tests__/unit/resolvers/reactionResolver.test.ts
+++ b/quotevote-backend/__tests__/unit/resolvers/reactionResolver.test.ts
@@ -22,7 +22,7 @@ describe('reactionResolver', () => {
   });
 
   describe('Query.actionReactions', () => {
-    it('returns reactions for a given actionId', async () => {
+    it('returns reactions for a given actionId using findByActionId', async () => {
       const mockReactions = [
         {
           _id: new mongoose.Types.ObjectId(),
@@ -32,7 +32,7 @@ describe('reactionResolver', () => {
         },
       ];
 
-      (Reaction.find as jest.Mock).mockReturnValue({
+      (Reaction.findByActionId as jest.Mock).mockReturnValue({
         lean: jest.fn().mockResolvedValue(mockReactions),
       });
 
@@ -41,7 +41,7 @@ describe('reactionResolver', () => {
         { actionId: 'action-123' }
       );
 
-      expect(Reaction.find).toHaveBeenCalledWith({ actionId: 'action-123' });
+      expect(Reaction.findByActionId).toHaveBeenCalledWith('action-123');
       expect(result).toHaveLength(1);
       expect(result[0].emoji).toBe('👍');
     });
@@ -58,7 +58,7 @@ describe('reactionResolver', () => {
       ).rejects.toThrow(/Authentication required/);
     });
 
-    it('creates and returns a new reaction', async () => {
+    it('upserts and returns reaction using context.user._id', async () => {
       const mockRxn = {
         _id: new mongoose.Types.ObjectId('60d5ec49ad414d7a8d5464b1'),
         userId: new mongoose.Types.ObjectId(userId),
@@ -67,11 +67,13 @@ describe('reactionResolver', () => {
         created: new Date(),
       };
 
-      (Reaction.create as jest.Mock).mockResolvedValue(mockRxn);
+      (Reaction.findOneAndUpdate as jest.Mock).mockReturnValue({
+        lean: jest.fn().mockResolvedValue(mockRxn),
+      });
 
       const result = await reactionResolver.Mutation.addActionReaction(
         null,
-        { reaction: { userId, actionId: '60d5ec49ad414d7a8d5464c2', emoji: '👍' } },
+        { reaction: { userId: 'other-user-id', actionId: '60d5ec49ad414d7a8d5464c2', emoji: '👍' } },
         mockContext({
           _id: userId,
           username: 'alice',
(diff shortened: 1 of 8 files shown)


motirebuma (changes requested):
Hey @AegisX-dev - the code base for the reaction system is looking good overall. The resolver structure, auth guards, and tests are all in order. A few security and polish items before merge.

#### What's good

1. Auth guards - both mutations check `context.user._id` and throw `UNAUTHENTICATED` with appropriate GraphQL error extensions. Tested both authenticated and unauthorized cases.

2. Frontend - optional chaining on `state.user.data?.id` prevents crash when unauthenticated, and `stopPropagation()` on the reaction container fixes the card expansion bug. Both are simple one-liners.

3. Tests - good coverage of query, create, update, and auth rejection scenarios.

#### Items to address

1. `addActionReaction` trusts client-provided `userId`. The mutation takes `userId` from `args.reaction.userId` and passes it to `Reaction.create()` directly. An authenticated user could add a reaction on another user's behalf. Use `context.user._id` instead:
```ts
const rxn = await Reaction.create({
userId: context.user._id, // insecure to take userId from args
actionId: args.reaction.actionId,
emoji: args.reaction.emoji
});
```

2. `updateActionReaction` needs ownership guard. Any authenticated user could update any reaction by `_id`. Before updating, verify that the reaction belongs to the current user:
```ts
const existing = await Reaction.findById(args._id).lean();
if (!existing || existing.userId.toString() !== context.user._id.toString()) {
throw new GraphQLError('Not authorized', { extensions: { code: 'FORBIDDEN' } });
}
```

3. Missing `deleteActionReaction` mutation. Currently, users can add and update reactions, but not delete them. Need a mutation to remove a reaction.

4. No guard in `addActionReaction` against multiple reactions from the same user. A user could click "React" multiple times and create multiple identical reactions. Consider using `findOneAndUpdate(..., { upsert: true })` on `(userId, actionId)` index, or add a unique index on these fields.

#### Nit

The `Reaction` model already defines a static `findByActionId` method. The query resolver should use `Reaction.findByActionId(args.actionId)` rather than `Reaction.find({ actionId: args.actionId })` - that way, if the implementation of `findByActionId()` changes in the future, the resolver will continue to work.

Thanks @AegisX-dev 

@flyblackbox @AegisX-dev

motirebuma (approved):
Hi @AegisX-dev, superb! All of my comments are addressed, and the implementation looks good.

### What is good

#### Security issues

`addActionReaction` makes use of `context.user._id` rather than the given `userId` in the arguments. `updateActionReaction`, as well as the new `deleteActionReaction`, now have ownership guarding, returning `FORBIDDEN` if the user is not the one who created the reaction.

#### Duplicate guard

The new `addActionReaction` uses `findOneAndUpdate(..., { upsert: true })` on `(userId, actionId)` combo, and a `{ userId: 1, actionId: 1 }` unique index on the model. A belt and suspenders approach.

#### Delete mutation

The new delete mutation has an auth guard, ownership check, a GraphQL schema mutation, and TypeScript type definition. The frontend also has the `DELETE_ACTION_REACTION` mutation.

#### Other type fixes

The static methods now return `QueryWithHelpers<...>` rather than `Promise`, so that `.lean()` can be chained on them. Good catch!

#### Model statics usage

Using `Reaction.findByActionId()` static method rather than the raw `find()`.

#### Test coverage

All the tests are present and cover the happy path, ownership check rejection (FORBIDDEN), not-found, delete success, and the upsert in `addActionReaction` correctly uses `context.user._id` even if another userId is passed in arguments.

#### Frontend fixes

Optional chaining on the user store and `stopPropagation()` on the reaction container.

No issues.

thansk @AegisX-dev 

@flyblackbox @AegisX-dev

motirebuma (approved):
Hi @AegisX-dev, Good job answering all previous review comments and implementing their suggestions. All looks good

**What's good**

1. All security issues addressed. The action reaction now uses context.user._id instead of a userId passed from the client. Ownership checks added to updateActionReaction and the new deleteActionReaction with FORBIDDEN code

2. The duplicate guard implemented. addActionReaction uses findOneAndUpdate with upsert: true on (userId, actionId) pair. Also, a new unique compound index {userId: 1, actionId: 1} added to the model. Belt and suspenders approach, as required

3. The new delete mutation added. Implemented with auth guard, ownership check, schema, TS type, and frontend DELETE_ACTION_REACTION mutation. All required parts present

4. Types fixed. Static methods return type is QueryWithHelpers<...> instead of Promise. Allows to call .lean() in the query

5. Used model statics. Reaction.findByActionId() is used instead of Reaction.find()

6. All necessary tests written. They check forbidden on ownership, not found, successful deletion, and the upsert approach, where context.user._id is used, even if another userId is passed in args

7. Frontend changes: optional chaining when accessing user store and stopPropagation() call on reaction container click. All looks good

No issues

**Approved**

Thanks @AegisX-dev 

@flyblackbox @AegisX-dev

flyblackbox:
Thank you @AegisX-dev!  Great work. Are you open to take on a new assignment?

AegisX-dev:
Thanks! @flyblackbox Glad you liked it. Yes, I'm open to it.

Issue #440, opened 2026-07-25
Profile page navigation error blocks profile acceptance testing


flyblackbox:
## Summary

A profile-page error currently prevents completion of acceptance testing for the About/bio editing work delivered in #434 and tracked in #362.

## Current evidence

@niamao reported that an error occurs when navigating to the profile page. The exact route, account state, visible error, and underlying console/network/GraphQL failure still need to be captured.

## Reproduction details needed

- [ ] Record the affected profile route.
- [ ] Record whether the viewer is authenticated, logged out, or viewing another user.
- [ ] Provide reliable reproduction steps.
- [ ] Capture the visible error state.
- [ ] Capture relevant console, network, and GraphQL errors.
- [ ] Confirm whether the `updateUser` bio mutation succeeds before navigation or rendering fails.

## Resolution tasks

- [ ] Identify the failing component, query, or route transition.
- [ ] Add regression coverage for the reproduced failure.
- [ ] Restore successful profile-page navigation and rendering.
- [ ] Re-test the About field save-and-persist flow from #362.

## Acceptance criteria

- The affected profile route loads without the reported error.
- Existing profile content renders for the relevant account state.
- About/bio changes can be saved, followed by a successful profile reload.
- A regression test covers the underlying failure.

## Related work

- Implementation: #434
- Acceptance testing: #362

flyblackbox:
Need to revalidate against current main. The report still says the exact route/error/GraphQL failure needs to be captured. Since then #439 landed profile/presence/API work and #446 corrected production GraphQL compatibility. @niamao or @motirebuma can you check this one?

niamao:
Revalidated on current main (#439 / #446) and hardened profile navigation.

**What was going wrong**
- `GET_USER` must not request `bio` (hosted `User` has no such field) or the whole profile query dies.
- Activity cards called `url.replace` when `post.url` was missing, which crashed the profile activity tab.
- Query failures were indistinguishable from “Invalid user”.

**What landed**
- Profile still loads via `GET_USER` even when About/bio is unavailable.
- About text is fetched separately (`GET_USER_BIO`) with `errorPolicy: 'all'`.
- Activity navigation no longer throws on missing post URLs.
- GraphQL errors show a recoverable “Couldn’t load this profile” state instead of a crash.
- Settings shows an About field only when the API actually supports `bio`.
- Regression tests cover the error UI, isolated bio query, and `toAppPostHref(null)`.

[...]

PR #456, opened 2026-08-17
feat(profile): restore About/bio behind a hosted-safe query

niamao:
## Summary
- Restore the Settings About field and profile About text by loading `bio` through a separate `GET_USER_BIO` query, so a hosted schema without `User.bio` cannot take down the profile page (#440 / #362).
- Save About only when the API actually supports `bio` (local `quotevote-backend`); keep the field hidden against `api.quote.vote`.
- Surface a recoverable profile error instead of “Invalid user” when `GET_USER` itself fails.

## Test plan
- [ ] Against local backend (`localhost:4000`): Settings shows About, save persists, profile About tab shows the text
- [ ] Against hosted API (`https://api.quote.vote`): Settings hides About, profile still loads, About tab shows empty state
- [ ] Profile GraphQL errors show “Couldn’t load this profile” with a Back to Explore link
- [ ] `pnpm test -- src/__tests__/graphql/queries.test.ts src/__tests__/components/Profile/ProfileController.test.tsx src/__tests__/components/Profile/ProfileView.test.tsx src/__tests__/app/dashboard/settings/page.test.tsx`
- [ ] `pnpm validate:graphql` still passes (GET_USER_BIO is optional)

### DIFF
--- a/quotevote-frontend/scripts/validate-graphql.mjs
+++ b/quotevote-frontend/scripts/validate-graphql.mjs
@@ -25,6 +25,10 @@ if (!fs.existsSync(schemaPath)) {
 
 const schema = buildSchema(fs.readFileSync(schemaPath, "utf8"));
 
+// Documents that probe local-only fields (e.g. User.bio). Callers use
+// errorPolicy: 'all' so hosted schemas reject them without taking down the page.
+const OPTIONAL_DOCUMENTS = new Set(["GET_USER_BIO"]);
+
 let checked = 0;
 const failures = [];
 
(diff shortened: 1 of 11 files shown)


flyblackbox (commented):
Looks good! Need final sign off from @motirebuma

motirebuma (approved):
Hey @niamao nice work. Loading the bio through a separate query so the hosted API can't break the whole profile page is the right call. If bio isn't available, the About field just quietly disappears, no crash, no blank page. Settings handles it the same way. Tests cover all the important cases (bio works, bio unavailable, empty bio, full query failure). The error state with "Back to Explore" is a nice touch too.

thank you @niamao 

LGTM

@flyblackbox @niamao

PR #462, opened 2026-08-19
fix: Neo Brutalism discussion header ink bar

niamao:
## Summary
- Stops Neo Brutalism from treating the mobile discussion resize handle as a menu separator, which drew a full-width black line through the yellow Discussion header ([#459](https://github.com/QuoteVote/quotevote-next/issues/459)).
- Tightens grabber padding so the yellow header reads as one block.

## Test plan
- [ ] Enable Neo Brutalism, open a post on portrait mobile, open Discussion
- [ ] Yellow `Discussion · N` header is a single block (no black line cutting through it)
- [ ] Divider still drags to resize Quote / Discussion panes
- [ ] Default (non–Neo Brutalism) theme is unchanged
- [ ] Dropdown/menu separators still render as ink rules under Neo Brutalism

Made with [Cursor](https://cursor.com)

PR #472, opened 2026-08-21
fix: enlarge mobile discussion resize hit area

niamao:
## Summary
- Follow-up to #462: the grabber pill was left at 4px tall, which is too small to drag on a phone.
- The whole discussion header is now the resize handle (pill stays a visual cue). The collapse chevron is excluded so it still closes instead of starting a drag.
- Default theme keeps a little top margin on the pill; Neo Brutalism drops it so the yellow header stays one tight block.

## Test plan
- [ ] Open a post on portrait mobile, expand Discussion
- [ ] Drag from the **Discussion · N** title (not just the pill) and confirm the Quote / Discussion split resizes
- [ ] Collapse chevron still closes Discussion and does not start a drag
- [ ] Default theme: small gap above the pill
- [ ] Neo Brutalism: no extra gap above the pill; yellow header is still one block (no ink bar)

### DIFF
--- a/quotevote-frontend/src/__tests__/components/Post/MobileDiscussionSplit.test.tsx
+++ b/quotevote-frontend/src/__tests__/components/Post/MobileDiscussionSplit.test.tsx
@@ -1,6 +1,23 @@
-import { render, screen, fireEvent } from "@/__tests__/utils/test-utils";
+import { createEvent, fireEvent, render, screen } from "@/__tests__/utils/test-utils";
 import MobileDiscussionSplit from "@/components/Post/MobileDiscussionSplit";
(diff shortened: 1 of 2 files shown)


motirebuma (approved):
Hey nice work, Follow-up to #462. The 4px pill was too small to grab on a phone, now the entire discussion header is the drag target. The pill stays as a visual hint but you can start dragging from anywhere on the header row.

Smart detail: the collapse chevron calls `e.stopPropagation()` on `pointerDown` so tapping it still closes the panel instead of starting a drag. Tests cover both the resize-from-header and the collapse-button-excluded cases.

The Neo Brutalism tweak (`.neo-brutalism &:mt-0`) keeps the yellow header tight while default theme gets a small top margin on the pill. Clean separation.

LGTM

thanks @niamao 

@flyblackbox @niamao

Issue #487, opened 2026-08-27
📌 Make Mobile Unauthenticated Nav and Discovery Controls Sticky


flyblackbox:
# 📌 Make Mobile Unauthenticated Nav and Discovery Controls Sticky

## Summary

On mobile unauthenticated pages, keep the nav and main discovery controls visible while the user scrolls through posts.

## Requirements

Make these elements sticky:

- Top navigation
- Search bar
- Filter row

The sticky area should remain fixed at the top of the viewport while the post list scrolls underneath it.

Preserve the current visual order:

1. Top navigation
2. Search
3. Filters
4. Post results

Avoid overlapping content or creating duplicate spacing when the sticky state activates.

## Acceptance Criteria

- [ ] Top navigation remains visible while scrolling.
- [ ] Search remains visible while scrolling.
- [ ] Filters remain visible while scrolling.
- [ ] Post cards scroll underneath the sticky controls.
- [ ] Sticky controls do not overlap one another or the post content.
- [ ] Behavior applies to mobile unauthenticated views.
- [ ] Desktop behavior remains unchanged.

## Labels

`enhancement` `mobile` `frontend` `UX` `unauthenticated`

PR #488, opened 2026-08-27
fix(directory): keep mobile guest nav and filters pinned while scrolling

niamao:
## Summary
- Pin the public directory nav, search, and filter row on mobile so they stay visible while posts scroll underneath ([#487](https://github.com/QuoteVote/quotevote-next/issues/487)).
- Use a mobile scroll shell instead of `position: sticky`, which cannot activate because `html`/`body` set `overflow-x: hidden`.
- Leave desktop page scrolling unchanged so the toolbar is not pinned.

## Test plan
- [ ] Open `/` as a guest on a mobile viewport (~390px).
- [ ] Confirm visual order is still: top nav, search, filters, then posts.
- [ ] Scroll the post list and confirm nav, search, and filters stay at the top with no overlap or extra gap.
- [ ] Open Filters / Tag / Date popovers after scrolling and confirm they still work.
- [ ] On desktop, confirm the toolbar still scrolls away with the page.
- [ ] `pnpm lint && pnpm type-check && pnpm test -- src/__tests__/app/page.test.tsx` in `quotevote-frontend/`.

### DIFF
--- a/quotevote-frontend/src/__tests__/app/page.test.tsx
+++ b/quotevote-frontend/src/__tests__/app/page.test.tsx
@@ -80,6 +80,23 @@ describe('Public directory (`/`)', () => {
     expect(screen.getByTestId('filter-date')).toBeInTheDocument()
   })
 
+  it('keeps nav, search, and filters pinned above the scrolling post list on mobile', () => {
+    renderDirectory()
+    const chrome = screen.getByTestId('directory-sticky-chrome')
+    const scroll = screen.getByTestId('directory-scroll')
+    expect(chrome).toContainElement(screen.getByRole('navigation', { name: 'Main navigation' }))
+    expect(chrome).toContainElement(screen.getByTestId('directory-toolbar'))
+    expect(chrome).toHaveClass('shrink-0')
+    expect(screen.getByTestId('public-directory')).toHaveClass(
+      'h-dvh',
+      'overflow-hidden',
+      'md:h-auto',
+      'md:overflow-visible'
+    )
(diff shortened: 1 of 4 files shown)


motirebuma (approved):
Hey @niamao Nice fix. The reasoning is solid -- `position: sticky` genuinely doesn't work when `overflow-x: hidden` is on the scroll ancestor, so wrapping the chrome in a flex `shrink-0` container and letting the post list scroll independently is the right call.

A few notes:

**Approach is clean.** Pinning the nav + toolbar above the scroll area on mobile while leaving desktop behavior untouched via `md:` breakpoint overrides is exactly the right scope. No extra JS, no resize listeners, just CSS.

**Test covers structure well.** Checking that the chrome wrapper contains the nav and toolbar, and that the scroll container does not, validates the layout contract without being fragile.

**One thing to double-check:** The `DirectoryToolbar` now has a hardcoded `bg-[#eef4f9]` background. This matches the parent container's inline `style={{ background: '#eef4f9' }}`, so it looks correct for preventing content from showing through during scroll. Just make sure this stays in sync if that background color ever changes -- might be worth extracting it to a shared constant at some point, but not a blocker for this PR.

LGTM

@flyblackbox @niamao

flyblackbox:
Also should apply to authenticated users as well.  Thank you @niamao for the call out!

niamao:
> Also should apply to authenticated users as well. Thank you @niamao for the call out!

addressed it here https://github.com/QuoteVote/quotevote-next/pull/494

Issue #473, opened 2026-08-21
Replace /Explore with root /


flyblackbox:
Delete the explore page, replace it with the / root page.

AegisX-dev:
i'll be working on this issue.

flyblackbox:
Thank you @AegisX-dev!  I have reassigned this one @niamao.

flyblackbox:
This one didn't deploy successfully.  Can you help me work through that @motirebuma? I will DM you

PR #491, opened 2026-08-27
fix(#490): auth-aware root page and remove /dashboard route prefix

### LINKED ISSUE #490: Authenticated users see guest UI after logging in (root / page has no auth awareness)
## Problem

After logging in, users are redirected to `/` but still see the guest header (with "Login", "Request Invite", "About") instead of the authenticated dashboard UI (nav bar with profile menu, sidebars, bottom tabs). The login technically works (the token is set), but the page doesn't reflect it.

## Root cause

PR #481 replaced `/dashboard/explore` with `/` as the home feed. Before that change, two things kept the authenticated experience working:

1. `PublicDirectoryContent` had a `useEffect` that checked `user?._id` and redirected logged-in users to `/dashboard/explore`
2. `/dashboard/explore` lived inside `src/app/dashboard/`, so it was wrapped by the dashboard layout which provides the authenticated nav, sidebars, and bottom tabs

PR #481 removed the redirect from `PublicDirectoryContent` (and `LandingPageContent`) so that `/` could serve as the unified home for both guests and logged-in users. But the root page (`src/app/page.tsx`) still renders `DirectoryHeader`, which is a hardcoded guest-only header with no auth state awareness.

The middleware (`middleware.ts`) also doesn't run on `/` since the matcher only covers `/dashboard/:path*` and `/auths/:path*`.

## What's affected

- `src/app/page.tsx` renders `PublicDirectoryContent` with no auth check
- `src/app/components/PublicDirectory/DirectoryHeader.tsx` always shows guest nav (Login, Request Invite, About)
- `src/app/components/PublicDirectory/PublicDirectoryContent.tsx` has no user/auth state
- Post-login redirect (`router.push('/')`) sends users to this guest-only page
- Post-signup redirect also goes to `/`

## Expected behavior

Logged-in users at `/` should see the authenticated UI: dashboard nav bar with profile menu, sidebars (notifications, messages), bottom tabs on mobile, and the post feed.

## Possible approaches

1. Make the root `/` page auth-aware: detect the user's login state and conditionally render the dashboard layout (nav, sidebars) for authenticated users and the guest `DirectoryHeader` for guests
2. Redirect authenticated users from `/` to a dashboard route that has the full layout

Option 1 aligns with the original intent of PR #481 (unified home for everyone) but requires the root page to conditionally use the dashboard shell. Option 2 is simpler but partially reverts the migration goal.

@flyblackbox

motirebuma:
## Summary

Closes #490 and addresses #473.

- **Auth-aware root page**: The `/` route now renders the authenticated dashboard shell (navbar, sidebars, bottom tabs, feed) for logged-in users and the public directory for guests. No redirect is involved; rendering is conditional via client-side auth state from the Zustand store.
- **Remove `/dashboard` prefix from all routes**: Every route that was previously under `/dashboard/...` now lives at the top level (`/profile`, `/settings`, `/notifications`, `/post`, `/control-panel`, `/manage-invites`). The old `/dashboard` layout is replaced by a Next.js route group `(dashboard)` that shares the same shell layout without affecting the URL.
- **Backward compatibility**: A catch-all page at `src/app/dashboard/[[...path]]/page.tsx` redirects any legacy `/dashboard/...` URL to its new equivalent. `/dashboard/explore` maps to `/`.
- **Middleware moved to `src/middleware.ts`**: Next.js 16 requires middleware inside the `src/` directory. Route protection and auth gates updated to match the new paths.

## Key changes

| Area | What changed |
|------|-------------|
| `src/app/page.tsx` | Renders `AuthAwareHome` instead of `PublicDirectoryContent` directly |
| `src/app/components/AuthAwareHome.tsx` | New client component: conditionally renders `DashboardShell` or public directory based on auth state |
| `src/components/DashboardShell.tsx` | Extracted from the old dashboard layout; reusable shell with navbar, sidebars, chat panel, presence hooks |
| `src/app/(dashboard)/` | Route group replacing `src/app/dashboard/` for all protected pages |
| `src/app/dashboard/[[...path]]/page.tsx` | Catch-all redirect for backward compatibility |
| `src/middleware.ts` | Moved from project root; updated protected prefixes and function imports |
| `src/lib/dashboard-routes.ts` | Functions renamed (`isGuestReadableRoute`, `isAuthRequiredRoute`); paths updated |
| `src/lib/utils/chatLayout.ts` | Persistent chat panel routes updated |
| `src/lib/utils/sanitizeUrl.ts` | `toAppPostUrl` simplified (no longer prepends `/dashboard`) |
| `next.config.ts` | Legacy redirects removed (were causing redirect loops) |
| 40+ component files | All internal links and navigation calls updated from `/dashboard/...` to `/...` |
| 28 test files | All test expectations updated to match new routes; new tests for catch-all redirect |

## Test plan

- [x] `pnpm type-check` passes with zero errors
- [x] All 164 test suites pass (1942 tests, 4 pre-existing skips)
- [x] Manual verification: new routes (`/settings`, `/profile/:user`, `/post/...`) work correctly
- [x] Manual verification: auth gates redirect unauthenticated users to `/auths/login`
- [x] Manual verification: guest-readable routes (`/post/...`, `/profile/:user`) accessible without login
- [x] Manual verification: legacy `/dashboard/...` URLs redirect to new paths
- [x] Root `/` shows public directory for guests, dashboard feed for authenticated users


motirebuma:
@flyblackbox This PR makes the root `/` page auth-aware (authenticated users see the dashboard shell, guests see the public directory) and removes the `/dashboard` prefix from all routes (`/dashboard/profile` → `/profile`, `/dashboard/settings` → `/settings`, etc.). Legacy `/dashboard/...` URLs are handled by a catch-all redirect for backward compatibility. All 1942 tests pass, type check is clean.

flyblackbox (approved):
Great work identifying this need and solving it @motirebuma.  Approved for merge!

motirebuma:
thanks @flyblackbox

PR #494, opened 2026-08-28
fix(directory): pin search and filters for signed-in mobile users

niamao:
## Summary
- Apply the #488 mobile scroll shell to authenticated `/` so search and filters stay pinned while posts scroll ([#488 comment](https://github.com/QuoteVote/quotevote-next/pull/488#issuecomment-followup)).
- Leave desktop page scrolling unchanged; `DashboardShell` still owns the fixed nav.
- Replace hardcoded `#eef4f9` with `bg-background` so the toolbar stays in sync with the page background.

## Test plan
- Open `/` signed in on a mobile viewport (~390px).
- Confirm visual order is: logo bar, search, filters, then posts.
- Scroll the post list and confirm search and filters stay at the top with no overlap or extra gap.
- Open Filters / Tag / Date popovers after scrolling and confirm they still work.
- On desktop, confirm the toolbar still scrolls away with the page.
- As a guest, confirm the existing pinned chrome is unchanged.
- `pnpm lint && pnpm type-check && pnpm test -- src/__tests__/app/page.test.tsx` in `quotevote-frontend/`.

### DIFF
--- a/quotevote-frontend/src/__tests__/app/page.test.tsx
+++ b/quotevote-frontend/src/__tests__/app/page.test.tsx
@@ -5,8 +5,10 @@
  * and no marketing landing sections.
  */
 
+import { type ReactNode } from 'react'
 import { render, screen } from '../utils/test-utils'
 import { PublicDirectoryContent } from '@/app/components/PublicDirectory/PublicDirectoryContent'
+import { AuthAwareHome } from '@/app/components/AuthAwareHome'
 import { useAppStore } from '@/store'
 
 const mockPush = jest.fn()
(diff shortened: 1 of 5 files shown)


flyblackbox (commented):
Looks good to me, up to @motirebuma for review and approval now!

motirebuma (approved):
Hey @niamao this is a solid follow-up to #488. Bringing the same pinned-chrome pattern to the authenticated feed makes the scrolling behavior consistent for both guests and signed-in users, which is how it should be.

Replacing the hardcoded `#eef4f9` with `bg-background` across both `PublicDirectoryContent` and `DirectoryToolbar` is a good cleanup -- that hex was duplicated in three places and would've drifted eventually.

The `DashboardShell` changes are well-scoped. Conditionally switching `overflow-hidden` vs `overflow-y-auto` on `<main>` based on `pathname === '/'` keeps the pinned behavior isolated to the home feed without touching other routes. And the `h-full min-h-0` wrapper inside `AuthenticatedFeed` gives flexbox the right constraints to calculate the scroll area.

Test mirrors the guest directory structure nicely, which makes it easy to verify both paths stay in sync.

CI is green. LGTM

@flyblackbox @niamao

PR #512, opened 2026-09-07
fix: comment deep links, vote switching, desktop messages

niamao:
## Summary
- Fix comment deep links that appended `/comment` (invalid App Router path → broken opens); copy `#commentId` hashes, redirect legacy URLs, and focus the shared comment on load
- Allow passage vote switching in `VotingPopup` when retract/change is supported (disagree / change-vote no longer no-ops)
- Expose Messages on the desktop navbar so messaging is reachable without mobile viewport mode

## Test plan
- [ ] Copy a comment link from a post discussion → open in a new tab → lands on the post and highlights/scrolls to that comment
- [ ] Open a legacy `/post/.../comment#id` URL → redirects to `/post/...#id` and still focuses the comment
- [ ] Select text → agree with a tag → select again → change to disagree via a downvote tag → vote updates
- [ ] On desktop (≥ md), click Messages in the top navbar → messages panel opens; unread badge still works
- [ ] `pnpm lint && pnpm type-check && pnpm build` in `quotevote-frontend/`


flyblackbox (commented):
Lgtm needs final review from @motirebuma

motirebuma (approved):
Hi @niamao Nice work on this one. The comment deep link fix is clean - replacing the broken /comment path segment with a hash-only approach and adding the toCommentDeepLink utility is the right call. The Next.js redirects for legacy URLs are a good touch too.

Vote switching fix looks correct - the hasVoted guard was too aggressive, blocking vote changes even when onDeleteVote was available. Adding onDeleteVote to the dependency array is the right fix.

Desktop messages button in the navbar makes sense. Tests cover all three changes.

LGTM

Thanks @niamao 

@flyblackbox @niamao

PR #513, opened 2026-09-08
fix: research bug hunt — auth, share, selection, UX, CSP

niamao:
## Summary
- Harden Next.js middleware: verify HS256 tokens when `JWT_SECRET` is set, never trust unsigned admin claims, and clear expired/invalid cookies so login remains reachable
- Share/copy always uses a canonical `/post/...` URL (or fails visibly) instead of falling back to the feed/profile page
- Stabilize desktop text selection voting: commit on pointerup without mid-drag jumps, keep the browser selection so the toolbar stays open
- Clarify Create Quote title hierarchy and profile activity multi-select filters (`aria-pressed` + clear)
- Drop `unsafe-eval` from production CSP

## Test plan
- [ ] Select post text on desktop → release drag → agree/disagree popup stays open with highlight intact
- [ ] Visit `/auths/login` with an expired `qv-token` cookie → login page loads (cookie cleared)
- [ ] With `JWT_SECRET` set, forged `admin: true` cookie cannot open `/control-panel`
- [ ] Share from a feed/profile post card → clipboard gets `/post/...` URL, not the current page
- [ ] Create Quote shows persistent Title label; profile filters stack and show Clear filters
- [ ] `pnpm lint && pnpm type-check && pnpm build` in `quotevote-frontend/`


motirebuma (changes requested):
Hi @niamao this covers a lot of ground - auth hardening, share URLs, selection stability, profile filters, and CSP. Some thoughts:

The JWT session module is well structured. Verifying HS256 signatures when JWT_SECRET is available and refusing to trust unsigned admin claims is the right security posture. The fallback path (no secret = structural + expiry checks only, admin never trusted) is sensible for dev environments. Clearing expired cookies so users can reach the login page is a good catch.

The CSP change is straightforward - dropping unsafe-eval in production and extracting the policy into a testable module. Having the csp.test.ts to lock this down is good practice.

Desktop selection committing on pointerup instead of mid-selectionchange should fix the page-jump issue from the Matt Racz research findings. The new commitDesktopSelection test helper makes the test intent clearer.

Profile activity filters moving from tabs (aria-selected) to toggle buttons (aria-pressed) with multi-select makes more sense for the filtering use case since users can combine filters. Clean change

Fix the conflict and it will be merged.

thanks @niamao 

@flyblackbox @niamao

niamao:
@flyblackbox @motirebuma merge conflicts resolved

motirebuma (approved):
Hi @niamao nice work

LGTM

@flyblackbox @niamao

Issue #147, opened 2025-11-28
7.39 Migrate Subscription Resolvers to TypeScript and Prisma


flyblackbox:
# 7.39 Migrate Subscription Resolvers to TypeScript and Prisma

## Summary

Restore and validate the GraphQL subscription runtime for the migrated TypeScript backend.

The current backend does not have a wired subscription resolver or WebSocket server path, so focus on establishing the actual subscription infrastructure rather than copying the legacy implementation file-by-file.

## Scope

- Identify real-time behavior that remains part of the current product.
- Implement and register required TypeScript subscription resolvers.
- Establish the GraphQL WebSocket subscription runtime and pub/sub implementation.
- Use a typed subscription context consistent with #153.
- Preserve filtering and delivery semantics for retained subscriptions.
- Validate message, presence, typing, and other currently supported real-time behavior.
- Use Prisma only where subscription operations require database access.
- Add automated tests and validate delivery between two connected clients.

## Roster/Buddy Decision

#508 retires the legacy Roster/Buddy domain. The Roster/Buddy subscription surface should therefore be **removed rather than migrated**.

- Do not implement a new Roster/Buddy subscription in the TypeScript runtime.
- Remove obsolete Roster subscription declarations/client operations as #508 reaches retirement.
- Do not publish Buddy request/accept/decline/remove events for a retired relationship model.
- Preserve real-time messaging, Presence, and Typing independently where they remain supported.
- Blocking introduced by #508 does not require a replacement subscription unless a concrete product requirement is identified separately.

## Migration Approach

Treat legacy subscription code as a behavioral reference for retained real-time capabilities. Retired capabilities should be removed from schema/client surfaces rather than reproduced.

## Out of Scope

- Broad query resolver migration tracked by #142.
- Broad mutation resolver migration tracked by #144.
- Relationship resolver migration tracked by #146.
- Recreating Roster/Buddy subscription behavior retired by #508.
- Converting subscription code to Prisma when no database access is required.
- Unrelated real-time feature expansion.

## Acceptance Criteria

- [ ] GraphQL subscriptions for retained capabilities are registered in the executable schema.
- [ ] The TypeScript backend exposes a working WebSocket subscription endpoint.
- [ ] Subscription connections use a typed context compatible with #153.
- [ ] Authentication is available where required.
- [ ] Pub/sub events can be published and delivered to appropriate subscribers.
- [ ] Message, Presence, and Typing behavior is validated where supported.
[...]

PR #518, opened 2026-09-12
feat(graphql): integrate Apollo Server with typed context and Prisma …

### LINKED ISSUE #153: 7.45 Integrate Apollo Server with Typed Context
# 7.45 Integrate Apollo Server with Typed Context

## Summary

Establish the canonical typed GraphQL context for the migrated TypeScript backend and make Prisma Client available to resolvers through that shared context.

## Scope

- Define the canonical `GraphQLContext` used by backend resolvers.
- Add Prisma Client to the typed GraphQL context.
- Establish an application-level Prisma Client lifecycle; do not create one client per request.
- Preserve authenticated user information, request/response access where required, pub/sub, DataLoader support where applicable, and request metadata such as `requestId`.
- Make context suitable for both HTTP and WebSocket GraphQL operations.
- Add/update representative tests.

## Prisma Client Lifecycle

Prisma Client should be initialized and managed at the application level and injected through GraphQL context. Resolver migrations under #142, #144, and #146 should consume this shared client rather than creating independent instances.

## Authentication

Preserve current authentication behavior. Migration of individual authentication/user data-access paths belongs with the appropriate domain migration.

## Out of Scope

- Migrating every resolver from Mongoose to Prisma.
- Rewriting domain-specific authorization logic.
- Implementing the full WebSocket/subscription runtime tracked by #147.
- Removing Mongoose before dependent migrations are complete.

## Acceptance Criteria

- [ ] `GraphQLContext` defines the canonical resolver context shape.
- [ ] Prisma Client is available through context with an application-level lifecycle.
- [ ] Existing authenticated-user context behavior is preserved.
- [ ] Required request, response, pub/sub, DataLoader, and metadata capabilities remain available where applicable.
- [ ] HTTP operations receive the expected typed context.
- [ ] Context is compatible with #147 WebSocket work.
- [ ] Representative tests, backend type-check, and build pass.

## Dependencies

None required to begin. This should complete before or alongside the first Prisma domain migration.

## Related Issues

- #142 — Query resolver Prisma migration umbrella
- #144 — Mutation resolver Prisma migration umbrella
- #146 — Relationship resolver Prisma migration
- #147 — Subscription resolver and WebSocket migration

## Labels

`backend` `graphql` `typescript` `prisma` `apollo` `phase7`

shubham-01-star:
…singleton (#153)

Establish canonical typed GraphQLContext and application-level Prisma Client lifecycle for the QuoteVote backend to prepare for domain resolver migrations.

- Add application-level PrismaClient singleton with globalThis caching in app/lib/prisma.ts to avoid connection exhaustion during dev reloads

- Update GraphQLContext contract in app/types/graphql.ts with prisma, userId, and export HttpGraphQLContext / WsGraphQLContext subtypes

- Extract context creation into testable createHttpContext factory in app/context.ts supporting dependency injection

- Integrate createHttpContext into Apollo Server expressMiddleware in app/server.ts with SIGINT/SIGTERM graceful shutdown hooks

- Update resolver mockContext helpers across existing unit/integration tests

- Add unit tests for context factory (__tests__/unit/context.test.ts)

- Add end-to-end integration tests (__tests__/integration/typedContext.integration.test.ts)

Resolves #153


niamao:
## Review

The typed context and shared Prisma Client look correct: singleton lifecycle, `createHttpContext` with DI, auth/`requireAuth` behavior preserved, and `userId` derived from the hydrated user rather than the JWT. Backend CI is green.

A few contract/runtime nits so later resolver migrations share one pattern:

1. **Document the injection path.** In `app/lib/prisma.ts`, document that GraphQL resolvers use `context.prisma`. `import { prisma } from '~/lib/prisma'` should be limited to process startup, shutdown, and scripts. Resolver-level imports skip the factory and make the client impossible to inject in tests.

2. **Tighten the context contract.** Make `user`, `userId`, and `requestId` required on `GraphQLContext` (`User | null`, `string | null`, `string`) to match what the factory always returns. Optional keys leave `undefined` vs `null` ambiguity for every consumer.

3. **Align test helpers.** Have `mockContext` helpers set `userId` from `user._id`, not only a stub `prisma`. Right now a mock can look authenticated on `user` and logged-out on `userId`.

4. **Load env before constructing Prisma.** Load dotenv before `new PrismaClient()` (`import 'dotenv/config'` first, or lazy-init). `prisma.ts` is imported before `dotenv.config()` in `server.ts`, so `DATABASE_URL` from `.env` may not be visible at construction time.

### Non-blocking

- Shutdown disconnects DB clients but does not stop Apollo/HTTP.
- `WsGraphQLContext` is types-only (fine until the subscription work).
- Auth hydration staying on Mongoose is in scope.

No new authz/authn issues from the context change. Thanks for extracting the factory — this is the right shared path for the Prisma resolver migrations.

@shubham-01-star @motirebuma @flyblackbox

flyblackbox:
Hello @shubham-01-star!
Great work and fantastic turn around time. Thank you so much for your contribution. 

Codex reviewed the PR and found two issues worth considering/addressing while we wait for @motirebuma to review.  

1. The shutdown handler can run twice and disconnect databases before requests finish, 

and 2. the WebSocket context is incompatible with the existing subscription resolver types.

Could you address those two issues and add regression tests before we adopt this as the shared migration pattern?

Our first dev to use it will be @niamao who has already provided some feedback and will complete the first implementation once it's ready.

motirebuma (changes requested):
Good work @shubham-01-star, and nice turnaround on a critical issue. This is the foundation for the entire 7.77 migration track so I went through everything line by line.

The overall design is right: Prisma singleton with globalThis caching, context factory extracted from server.ts with DI support, and thorough test coverage at both unit and integration level. The requestId generation and userId convenience field are good additions that weren't in the original inline context.

I agree with the feedback from @niamao and @flyblackbox. Here's what needs to be addressed before merge:

**Must fix:**

1. Dotenv load order. In the new server.ts, `import { createHttpContext } from './context'` triggers `import { prisma } from './lib/prisma'` which calls `new PrismaClient()` at module load time. This happens before `dotenv.config()` on line 22, so `DATABASE_URL` from `.env` won't be set when the client is constructed. Fix: either add `import 'dotenv/config'` as the first line in prisma.ts, or lazy-initialize the client with a getter function.

2. Shutdown handler runs twice. If the process receives SIGINT followed by SIGTERM (common with container orchestrators), the shutdown function fires twice with no guard. Add a `let shuttingDown = false` flag at the top and return early on the second call. Also wrap the cleanup in try/catch so a failure in `disconnectPrisma()` doesn't skip `mongoose.disconnect()`. Simplest fix:

```ts
let shuttingDown = false;
const shutdown = async () => {
  if (shuttingDown) return;
  shuttingDown = true;
  console.log('Shutting down gracefully...');
  await Promise.allSettled([
    disconnectPrisma(),
    mongoose.disconnect(),
  ]);
  process.exit(0);
};
```

3. Make `user`, `userId`, and `requestId` required on GraphQLContext (as `User | null`, `string | null`, `string`). The factory always returns concrete values for all three, so keeping them optional creates undefined vs null ambiguity that every consumer has to handle. This is the right time to tighten the contract since every resolver will depend on it.

4. Test mockContext helpers should set `userId` derived from `user._id` to match the real factory behavior. Right now a mock can look authenticated via `user` but have no `userId`, which could hide bugs in resolvers that check `context.userId` instead of `context.user._id`.

**Non-blocking (better to have):**

5. The shutdown handler calls `process.exit(0)` without closing `httpServer` or calling `server.stop()` on Apollo. In-flight requests get killed rather than drained. Not blocking for merge since the old code had no shutdown handler at all, but worth adding in a follow-up.

6. WsGraphQLContext is types-only which is fine for now. Note that `SubscriptionResolver` uses `ResolverFn` which takes `GraphQLContext` (with `req`/`res`), so when the subscription work starts (#147), the resolver types will need to be generified over the context type or WsGraphQLContext will need to be unified. Not this PR's problem.

7. `HttpGraphQLContext` type alias is defined but not used anywhere. It's harmless as documentation but could be removed to reduce surface area. Up to you.

Backend CI, Frontend CI, and Frontend E2E are all green. Once the fix items are addressed this is good to merge. Great foundation for the domain migrations.

@flyblackbox @shubham-01-star @niamao

motirebuma (approved):
Good work @shubham-01-star. This is the foundation for the entire 7.77 migration track so I reviewed it carefully across both commits.

The core design is solid. The Prisma singleton in lib/prisma.ts uses the standard globalThis pattern for hot-reload safety, and importing dotenv/config before PrismaClient construction ensures DATABASE_URL is always available regardless of import order. The injection path documentation making it clear that resolvers must use context.prisma (not direct imports) will save contributors from a common testability mistake during the domain migrations.

The context factory in context.ts is a clean extraction from server.ts. It preserves all existing behavior (Bearer token extraction, user hydration via Mongoose, requireAuth enforcement, introspection bypass) and adds two useful fields: userId derived from user._id and requestId from x-request-id header or crypto.randomUUID(). The DI via ContextFactoryOptions means tests can swap in mock Prisma/PubSub without touching globalThis.

The tightened GraphQLContext contract is the right call. Making user (User | null), userId (string | null), and requestId (string) all required instead of optional eliminates the undefined vs null ambiguity that every resolver consumer would otherwise have to handle. Only loaders stays optional which makes sense since it's not wired up yet.

The shutdown handler is well done: double-signal guard with the shuttingDown flag, server.stop() to drain Apollo, httpServer.close() to stop accepting connections, and Promise.allSettled for the DB disconnects so one failure doesn't skip the other.

Test coverage is thorough. The unit tests cover the full context contract (Prisma lifecycle, pubsub injection, auth flow, requireAuth enforcement, introspection bypass) and the integration tests spin up a real Apollo Server to verify context reaches resolvers correctly, including Prisma instance identity across requests. All existing resolver test helpers now derive userId from user._id which keeps mocks aligned with the real factory.

The second commit cleanly addresses all feedback from @niamao and @flyblackbox. Good responsiveness.

LGTM

@flyblackbox @shubham-01-star @niamao

PR #522, opened 2026-09-16
feat: simplify post header and replace Agree/Disagree with Approve/Reject

### LINKED ISSUE #521: 🧭 Simplify and Reorganize the Post Header and Action UI
## Summary

Redesign the post header and primary action area to create a clearer hierarchy between **author information, post metadata, whole-post judgment, and passage-level reactions**.

Use the provided mockup as the visual reference for the intended layout.

This issue consolidates and supersedes #514, #519, and #520.

image

### Author

Keep the author information together:

- Avatar
- Display name / username
- **Follow / Following** button

Use an explicit text Follow/Following button rather than the person-plus icon.

Remove the **OP** badge.

### Post Metadata

Place post information on a separate row beneath the author:

- Community/tag displayed as `# community-name`
- Source/citation
- Publication date/time

The timestamp should be associated with the post rather than the author.

**Implementation note:** the current post-detail `Post.tsx` does not render community/tag metadata in this header. Verify the existing post/group data shape and reuse the canonical community/tag source rather than introducing duplicate state or mock-only data.

### Title

Give the post title strong visual priority beneath the metadata.

## Primary Post Actions

Use one consolidated action row:

**Reject [count] | Approve [count] | Save | Discussion [count] | •••**

### Approve / Reject

Replace whole-post **Agree / Disagree** with:

- 👎 **Reject**
- 👍 **Approve**

Display each aggregate count inside its corresponding button.

The user's current judgment should be visually identifiable, and users should be able to change their judgment.

**Approve and Reject must be mutually exclusive.** A user must not be represented in both `approvedBy` and `rejectedBy` at the same time. Verify whether this is already enforced by the backend; if not, ensure changing judgment removes the previous whole-post judgment before applying the new one.

Preserve existing whole-post voting data where technically compatible.

### Remove Duplicate Voting Controls

Remove the existing:

- Green up-arrow control
- Red down-arrow control
- Separate up/down aggregate counters

Approve/Reject should become the single representation of whole-post judgment.

### Other Actions

Keep readily accessible:

- **Save / Bookmark**
- **Discussion**, including discussion count

The number beside the Discussion control should represent **discussion activity (comments/replies)** rather than the current aggregate interaction count that also includes passage votes/quotes. Passage-level voting already has its own representation and should not inflate the speech-bubble count.

Place secondary actions in the `•••` overflow menu where applicable:

- Copy link
- Report
- Edit
- Delete

## Passage-Level Reactions

Do not change passage-level reactions.

Selected text should continue to support qualifiers such as:

- Agree
- Disagree
- True
- Like
- Future passage-level qualifiers

**Approve/Reject applies to the whole post; passage qualifiers apply to selected excerpts.**

## Responsive Behavior

Preserve the same conceptual hierarchy across desktop, tablet, and mobile:

**Author → Metadata → Title → Actions → Content**

Controls may adapt or wrap as space decreases, but their semantic grouping should remain intact.

## Acceptance Criteria

- [ ] Author identity and Follow/Following are visually grouped.
- [ ] Follow/Following uses an explicit text label.
- [ ] OP badge is removed.
- [ ] Community/tag, source, and timestamp form a distinct metadata row.
- [ ] Community/tag metadata uses the canonical existing post/group data source.
- [ ] Post title has greater visual priority than metadata and controls.
- [ ] Whole-post Agree/Disagree is replaced by Approve/Reject.
- [ ] Approve/Reject aggregate totals appear inside their respective buttons.
- [ ] The user's current whole-post judgment is identifiable and can be changed.
- [ ] Approve and Reject are mutually exclusive for each user.
- [ ] Existing whole-post voting data is preserved or migrated where appropriate.
- [ ] Legacy green/red arrow voting controls and separate counters are removed.
- [ ] Save/Bookmark remains directly accessible.
- [ ] Discussion remains directly accessible and its count represents comments/replies rather than passage votes or unrelated interactions.
- [ ] Secondary post actions are consolidated into the overflow menu.
- [ ] Passage-level reactions and their data remain unaffected.
- [ ] No duplicate whole-post voting interface remains.
- [ ] Accessibility labels reflect the updated controls.
- [ ] Layout remains clear across desktop, tablet, and mobile.

## Supersedes

- #514
- #519
- #520

niamao:
## Summary
- Redesign the post-detail header to group author identity with a text Follow/Following control, then community/source/timestamp, then a stronger title.
- Replace whole-post Agree/Disagree with mutually exclusive Approve/Reject counts, keep Save and a comment-only discussion count on the same row, and move copy/report/delete into the overflow menu.
- Leave passage-level reactions unchanged and reuse existing `groupId` / `GET_GROUP` / `approvedBy` / `rejectedBy` data.

Closes #521

## Test plan
- [ ] Open a post you do not own: Follow sits next to the display name, OP badge is gone, timestamp is in the metadata row
- [ ] Confirm community/tag uses the real group title and source uses `citationUrl` when present
- [ ] Approve, then Reject: only one is selected and counts update inside the buttons
- [ ] Confirm green/red arrow counters are gone and the speech-bubble count is comments only
- [ ] Save still works; overflow menu has Copy link, Report, and Delete (owner/admin)
- [ ] Select passage text: Agree/Disagree/True/Like popup is unchanged
- [ ] Check desktop and mobile: Author → Metadata → Title → Actions → Content


motirebuma (changes requested):
Nice work @niamao. The Approve/Reject redesign and optimistic update pattern are well done. A few things to address before merging.

**What works well:**

The optimistic update logic in `handleApprove` / `handleReject` is solid. The `prevApprovedBy` / `prevRejectedBy` pattern to sync local state with prop changes is the correct React approach for derived state. Mutual exclusivity (approving removes rejection and vice versa) is handled both optimistically and after server response. Error rollback resets to the current prop values which keeps the UI honest on failure.

The accessibility additions are good: `aria-pressed` on Approve/Reject, descriptive `aria-label` on Save and comment count, and moving destructive actions behind a DropdownMenu so they're not one mis-click away.

Tests are thorough. The new cases for Follow grouping, own-post hiding, metadata row, comment-only counts, and overflow menu placement all match the new behavior correctly.

**Must fix:**

1. **Hardcoded colors will break dark mode.** There are 25+ hardcoded hex values across Post.tsx, ApproveButton, and RejectButton (`#374151`, `#111827`, `#6b7280`, `#d1d5db`, `bg-white`, `#e9f8ef`, `#fdecec`, etc.). Several of these replace theme-aware tokens that worked before. For example, `text-foreground/85` became `text-[#374151]` for the post body, and `text-foreground` became `text-[#111827]` for the title. These are regressions - a user in dark mode will get dark text on a dark background. At minimum, use Tailwind's semantic tokens (`text-foreground`, `text-muted-foreground`, `bg-background`, `border-border`) for the neutral colors, and use `dark:` variants for the accent colors (approve green, reject red, source/community pills). The action bar elements all use `bg-white` which also breaks in dark mode.

2. **Duplicated pill styling.** The pattern `h-8 rounded-full px-2.5 text-xs font-semibold` appears on the Approve button, Reject button, BookmarkIconButton (showLabel variant), comment count element, and the overflow trigger. Extract a shared class (even a `cn()` constant in the file) so future styling changes only need one edit.

3. **GET_GROUP fires a separate query per post view.** `useQuery(GET_GROUP, { variables: { id: post.group } })` makes an extra network call to fetch the community title. If the parent query that loads the post could include `group { _id title }` as a nested field, that would eliminate the round-trip. If that's not feasible right now, at least confirm Apollo's cache covers repeated views so this doesn't fire on every re-render.

**Nit:**

4. **Quote formatting changes inflate the diff.** The single-to-double-quote changes in the test files are pure formatting noise (+616/-522 but the functional delta is much smaller). If this came from a Prettier/ESLint config change, commit the config + formatting separately so reviewers can focus on the real changes. Not blocking but worth doing for future PRs.

All CI checks pass (Backend CI, Frontend CI, Frontend E2E all green). The functional changes are good - fix the dark mode colors and this is ready to merge

niamao:
Thanks @motirebuma — addressed the must-fix items in 580140b.

**Dark mode:** Neutrals now use theme tokens (`text-foreground`, `text-muted-foreground`, `bg-background`, `border-border`) instead of hex/`bg-white`. Approve/Reject, Follow, source, and Save use `dark:` accent variants. Title and body are readable in dark mode again.

**Shared pill class:** Extracted `POST_ACTION_PILL_CLASS` and applied it to Approve, Reject, Save, and the comment count.

**GET_GROUP:** Nesting `group { _id title }` on the post isn't available right now — `PostType` only exposes `groupId`. The query already uses `cache-first` (same operation as `PostCard`), so a given `groupId` is served from Apollo cache on re-render and return visits in the session. Happy to add a nested `group` field in a follow-up if we want to drop the extra first-view request.

Quote-formatting nit noted for future PRs — leaving this diff as-is so this review stays focused.

motirebuma (approved):
Thanks @niamao, the second commit addresses everything cleanly.

Dark mode is solid now. All the hardcoded hex values are replaced with theme tokens (`text-foreground`, `text-muted-foreground`, `bg-background`, `border-border`) for the neutrals, and proper Tailwind utilities with `dark:` variants for the accent colors (green/red/blue/amber). The post body text is back to `text-foreground/85` which was the original theme-aware value. Good work going through every element systematically.

The `POST_ACTION_PILL_CLASS` extraction in `postActions.ts` is the right approach. Clean single source of truth for the shared pill chrome across Approve, Reject, Save, and comment count.

The GET_GROUP explanation makes sense. `cache-first` reuse with PostCard means a given group only fetches once per session. If you add a nested `group` field to `PostType` later that would be a nice cleanup but it's not blocking.

All CI checks pass (Backend CI, Frontend CI, Frontend E2E). Approving, good to merge.

LGTM

@flyblackbox @niamao

Issue #507, opened 2026-09-06
7.77.7 Migrate Group, Domain, and Collection Resolvers to Prisma


flyblackbox:
# 7.77.7 Migrate Group, Domain, and Collection Resolvers to Prisma

## Summary

Migrate existing Group, Domain, and Collection GraphQL query data access from Mongoose to Prisma while preserving current behavior.

The current resolver implementation is query-only. This issue does not add missing Mutation handlers. Contributes to #142 / 7.34 and is tracked under #500 / 7.77.

## Coordination

The Group domain overlaps with #429 (Rename Group to Tag). #429 should be completed or appropriately sequenced first so contributors do not migrate terminology that will immediately be replaced.

## Tasks

- [ ] Confirm #429 sequencing.
- [ ] Identify Mongoose-backed Group, Domain, and Collection query paths.
- [ ] Replace Mongoose model access with Prisma.
- [ ] Preserve relationships, authorization, filtering, and validation.
- [ ] Remove Mongoose imports from migrated paths.
- [ ] Add or update tests.

## Acceptance Criteria

- [ ] Existing Group/Tag, Domain, and Collection query data access uses Prisma.
- [ ] Existing records and relationships remain functional.
- [ ] No new Mutation handlers are introduced as part of this migration.
- [ ] Work is compatible with #429 sequencing.
- [ ] Relevant tests, type-check, and build pass.

## Dependencies

- #429 — Rename Group to Tag should be completed or appropriately sequenced first.
- #153 — canonical GraphQL context and Prisma Client lifecycle/injection pattern.

## Related Issues

- #142 — Query resolver Prisma migration umbrella
- #500 — Domain resolver Prisma migration umbrella

## Labels

`backend` `graphql` `typescript` `prisma` `phase7` `🟢 Parallelizable`

flyblackbox:
@gunjanghate would you like to take this one?

gunjanghate:
Yes will take on this one @flyblackbox

flyblackbox:
Amazing thank you.  
You are assigned! 🥳

PR #526, opened 2026-09-18
feat: Migrate 'Group' terminology to 'Tag' across monorepo (#429)

### LINKED ISSUE #429: Rename 'Group' to 'Tag'
# Rename 'Group' to 'Tag'

## Summary

Rename the post-classification concept currently called `Group` to `Tag` across the active Quote.Vote v2 frontend, GraphQL API, and backend.

The Create Quote interface already uses `Tag`, while the API and backend still use fields such as `groupId`, `createGroup`, and `GroupModel`.

A Tag is a searchable classification attached to a post. It has no membership, permissions, or public/private behavior. All posts are publicly visible.

## Tasks

- [ ] Find references to the post-classification concept using:
  - `Group`
  - `group`
  - `groupId`
  - `groups`
  - `createGroup`
  - `postTagId`
- [ ] Rename applicable GraphQL fields and operations:
  - `Group` → `Tag`
  - `groupId` → `tagId`
  - `group` → `tag`
  - `groups` → `tags`
  - `createGroup` → `createTag`
- [ ] Rename applicable backend models, types, resolvers, variables, and files
- [ ] Update the Next.js frontend to submit `tagId` directly
- [ ] Remove the `postTagId → groupId` mapping
- [ ] Preserve existing records and post relationships
- [ ] Update fixtures, tests, validation messages, and documentation
- [ ] Leave unrelated uses of `Group` unchanged

## Expected Behavior

The active application consistently uses `Tag` for the classification attached to a post.

Users can:

- create a Tag
- select an existing Tag
- submit a post with a Tag
- find posts by Tag
- open post URLs containing Tag metadata

Existing posts and Tags remain accessible.

## Acceptance Criteria

- [ ] `Tag` is the canonical term for post classification
- [ ] GraphQL uses tag-based fields and operations
- [ ] Backend application code uses Tag terminology
- [ ] Existing records and relationships remain intact
- [ ] Post creation and Tag-related tests pass
- [ ] Unrelated group mentions are not renamed
- [ ] Frontend and backend builds pass

## Labels

`backend` `graphql` `migration` `domain-model`

gunjanghate:
## Description

This PR completes the terminology migration from `Group` to `Tag` across the monorepo to ensure the frontend and backend codebases consistently use `Tag` as the canonical post-classification concept. 

To prevent downtime and ensure zero data loss, the underlying database collection name remains `groups`. This is strictly a model and codebase terminology update.

Closes #429

### Backend
* **Prisma**: Renamed `group.prisma` to `tag.prisma` and updated the model to `Tag`, using `@@map("groups")` to retain the existing database collection. Replaced all relations and foreign keys (e.g., `groupId` → `tagId`).
* **Mongoose**: Renamed `Group.ts` to `Tag.ts` and explicitly tied it to the `'groups'` collection. Updated references in `Post.ts` and `User.ts`.
* **GraphQL & Resolvers**: 
  * Renamed `Group` and `GroupInput` to `Tag` and `TagInput` in the type definitions.
  * Renamed `createGroup` mutation to `createTag`.
  * Updated resolver implementations (`tagResolver.ts`, `postsResolver.ts`) to handle `tag` and `tagId`.

### Frontend
* **GraphQL Sync**: Updated `schema.graphql` and client-side queries/mutations (`CREATE_TAG`, `TAGS_QUERY`, `GET_TAG`).
* **Component Logic**: Updated post submission logic in `SubmitPostForm.tsx` to natively submit `tagId` (removed the legacy `postTagId -> groupId` workaround).
* **Routing**: Renamed the dynamic post directory from `app/(dashboard)/post/[group]` to `app/(dashboard)/post/[tag]`.
* **Types & UI**: Replaced `Group` types and variables with `Tag` across UI components (e.g., `DirectoryToolbar`, `PaginatedPostsList`).

## Testing & Verification
* [x] Verified `pnpm exec tsc --noEmit` passes with 0 errors on both frontend and backend.
* [x] Updated backend unit tests and integration tests (`Tag.prisma.test.ts`, `Tag.schema.test.ts`) to mock and query tags.
* [x] Updated Playwright E2E fixtures to use `tagId` instead of `groupId`.
* [x] Ensured unrelated uses of the word "group" (such as Next.js route groups and chat/messaging rooms) were intentionally excluded from this migration and remain fully functional.


motirebuma (changes requested):
Good work @gunjanghate. The scope is well-contained and the rename is consistent across 70 files. The Tag model correctly uses `@@map("groups")` to keep the existing database collection, and `mongoose.model('Tag', TagSchema, 'groups')` does the same on the Mongoose side. The GraphQL schema, resolvers, types, queries, mutations, frontend routes, and tests are all aligned. CI is fully green.

Two critical issues before this can merge.

**Must fix:**

1. **Missing `@map` on renamed Prisma fields will break reads of existing data.** The model-level `@@map("groups")` is correct, but the field-level renames need `@map` too:

   - `post.prisma`: `tagId` needs `@map("groupId")` so Prisma reads the existing `groupId` field from stored post documents. Without it, every existing post returns `tagId: null`.
   - `user.prisma`: `memberOfTags` needs `@map("memberOfGroups")` for the same reason.

   CI passes because tests mock Prisma or create fresh documents with the new field names, so they never hit the real database mapping.

   Fix:
   ```prisma
   // post.prisma
   tagId  String  @map("groupId") @db.ObjectId

   // user.prisma
   memberOfTags  String[]  @map("memberOfGroups") @db.ObjectId
   ```

2. **Mongoose Post schema has the same field-level problem.** `Post.ts` renames `groupId` to `tagId` in the schema definition, but existing MongoDB documents store the field as `groupId`. Mongoose will read `tagId` from documents that don't have it (returning `undefined`) and write new posts with `tagId` while old posts still have `groupId`. You need to either keep the Mongoose field name as `groupId` (with an alias if you want `tagId` on the code side), or run a data migration to rename the field in all existing documents. Since this PR is specifically about zero-downtime renaming, keeping the Mongoose field as `groupId` with an alias is the safer approach.

3. **PR description has a placeholder instead of the issue number.** The body says "Addresses #[Insert Issue Number Here]". Update this to "Closes #429" so GitHub auto-links and auto-closes the issue on merge.

4. **Stale `groupId` in requireAuth test.** In the test query string, the argument reads `groupId` but should be `tagId` to match the updated schema. The test passes because requireAuth only checks the operation name, not the arguments, but it's inconsistent.

**Nit:**

5. **Index name changes create orphaned MongoDB indexes.** The renamed indexes (`tagId_1_featuredSlot_1` replacing `groupId_1_featuredSlot_1`, etc.) mean MongoDB will create new indexes on deploy while the old ones remain. After deploying, drop the old `groupId`-based indexes manually or via a migration script to avoid index bloat.

Fix the `@map` directives and the Mongoose field mapping, and this is ready to merge. The rest of the rename is thorough and consistent.

@flyblackbox @gunjanghate

gunjanghate:
@motirebuma 
All set! Added the missing `@map` directives in Prisma, reverted the Mongoose schema field back to `groupId` with a `tagId` alias, and fixed the stale `requireAuth` test. The underlying database structure and index names remain 100% untouched. Check once!

flyblackbox:
Thanks for turning around the review changes so quickly, @gunjanghate. We have a bit of a review backlog at the moment, but Moti will come back through this soon.

The compatibility work here is especially important because we want the terminology change without changing how the existing MongoDB data is stored. Keeping the Prisma fields mapped to groupId / memberOfGroups and retaining groupId underneath the Mongoose tagId alias looks aligned with that goal. Given the size of this PR and its role in unblocking #507, I’d hold off on additional scope for now and let @motirebuma verify the revised persistence mappings and the remaining review findings before merge. Thanks for being careful with the zero-downtime aspect of this.

motirebuma (changes requested):
Hi @gunjanghate  sorry for delayed review, and 4 of 5 from last round are sorted: the Prisma `@map("groupId")` / `@map("memberOfGroups")` are in, the requireAuth test and the PR body are fixed, and the index names went back to `groupId`, so no orphaned indexes. The Mongoose one is only half fixed: the `tagId` alias works on hydrated docs, but `postsResolver` and `featuredPostsResolver` read through `.lean()` and filter on `tagId`, so stored posts will throw or not match (details inline on `Post.ts:7`). I missed a bigger one last round: the frontend now calls `tags`, `tag(tagId)`, `createTag` and sends `tagId` in `PostInput`, but the hosted API is still the legacy server, which only serves `groups`, `group(groupId)`, `createGroup` and `groupId: String!`, and `schema.graphql` (a snapshot of api.quote.vote) was hand-edited in be33fb2 so `validate:graphql` stops catching that. Unless the v2 frontend already points at quotevote-backend in production, the frontend half needs to wait until the backend with `Tag` is the deployed API, so either split it into a follow-up PR or keep the frontend on the group fields for now, and revert the snapshot edit in both cases. Also agree with Copilot's note on `SubmitPost.tsx:39`: it reads `data.Tags` while `TAGS_QUERY` returns `tags` (base read `data.groups`), so the composer's tag list is always empty. this will be merged after you sorted out this, thanks @gunjanghate 

@flyblackbox @gunjanghate

gunjanghate:
@motirebuma 
I've updated both `postsResolver` and `featuredPostsResolver` to correctly query against and map `groupId` when using `.lean()` so they successfully match existing stored posts.

Also, I completely agree regarding the frontend/API rollout mismatch. I've entirely reverted the `quotevote-frontend` directory back to its state on `main` via `git restore` (which naturally resolves the `data.Tags` typo as well).

motirebuma (approved):
Everything from last round is sorted in e78adfc: `buildPostFilter` filters on `groupId` (postsResolver.ts:106), both lean mappers read `post.groupId` (postsResolver.ts:178, featuredPostsResolver.ts:53), the test fixture now looks like real lean output, and `quotevote-frontend` is back to exactly what's on main, so the snapshot edit and the `data.Tags` bug are gone too. Backend and Frontend CI are green on e78adfc. One thing to track, not a blocker: quotevote-backend now serves `tags`/`tag(tagId)`/`createTag`/`tagId` while the frontend on main still sends the group names, so running the frontend against a local backend will break tag features until the frontend half lands. @flyblackbox can we open a follow-up issue for the frontend rename, gated on quotevote-backend being the deployed API? Thanks for splitting it cleanly @gunjanghate.

@flyblackbox @gunjanghate

flyblackbox:
@motirebuma created #539 to track the frontend Group → Tag rename you asked us to split out, gated on `quotevote-backend` becoming the deployed API. I also asked you and @gunjanghate to review the new issue before we treat its scope as settled.

Issue #443, opened 2026-08-01
Implement Backend Support for `reportPost` and `reportBot` GraphQL Mutations


flyblackbox:
# Implement Backend Support for `reportPost` and `reportBot` GraphQL Mutations

## Summary

Implement the backend resolver logic for the existing `reportPost` and `reportBot` GraphQL mutations.

Both mutations are already declared in the GraphQL schema. The remaining work is to implement their resolver behavior, preserve the intended reporting workflow, and add automated coverage.

This backend work is a prerequisite for completing the reporting E2E coverage in #394 and #395.

## Current State

- `reportPost` is declared in the GraphQL schema.
- `reportBot` is declared in the GraphQL schema.
- Backend resolver implementations are still required.
- #394 and #395 depend on these mutations functioning correctly.

## Scope

- Implement the `reportPost` mutation resolver.
- Implement the `reportBot` mutation resolver.
- Follow the current backend resolver architecture and authentication patterns.
- Validate required mutation inputs.
- Persist reports using the appropriate existing data model.
- Preserve applicable authorization and validation behavior.
- Handle invalid targets appropriately.
- Return GraphQL responses consistent with the existing schema.
- Add automated tests for successful and unsuccessful reporting operations.

## Out of Scope

- Adding `reportPost` or `reportBot` to the GraphQL schema; both are already declared.
- Frontend reporting UI implementation.
- E2E implementation tracked by #394 and #395.
- Redesigning the moderation or reporting system.
- Unrelated Prisma migration work.

## Acceptance Criteria

- [ ] `reportPost` has a working backend resolver.
- [ ] `reportBot` has a working backend resolver.
- [ ] Existing GraphQL mutation definitions remain compatible.
- [ ] Required authentication and input validation are enforced.
- [ ] Reports are persisted through the appropriate backend data path.
- [ ] Invalid or unauthorized requests return appropriate GraphQL errors.
- [ ] Automated tests cover representative success and failure cases.
- [ ] GraphQL introspection exposes both mutations.
- [ ] Relevant backend tests and TypeScript type-check pass.
- [ ] #394 and #395 are unblocked from the backend side.

## Test Instructions

- Run the backend TypeScript type-check.
- Run the relevant resolver tests.
- Execute `reportPost` against a valid post.
- Execute `reportBot` against a valid bot target.
- Verify the expected report data is persisted.
- Verify invalid targets are handled correctly.
- Verify unauthenticated or unauthorized reporting behavior.
[...]

PR #527, opened 2026-09-19
Backend support for the reportPost and reportBot GraphQL mutations

sandy4242:
issue link - #443


flyblackbox:
Thanks for picking this up, @sandy4242. We’ve had several substantial PRs arrive at once, so reviews are running a little behind, but Moti will take a look soon.

While you’re waiting, one small thing that would make the review easier is expanding the PR description beyond issue link - #443: a short summary of how reportPost and reportBot are implemented, the files/paths changed, authorization or validation behavior, and the tests you ran would give @motirebuma a much faster starting point. Since reporting is a moderation boundary, it would also be helpful to call out explicitly how invalid targets, unauthenticated requests, duplicate reports, and reporter identity are handled if those cases are covered. 

Thanks for moving this long-standing backend gap forward. 🙏

motirebuma (changes requested):
Hi @sandy4242 the authorization is right on both mutations (the caller has to be the `userId`/`reporterId` they send, and self-reports are rejected), and the tests hit each failure path. Blocking: this branch is 112 commits behind main, and #525 has since moved `userResolver.ts` to Prisma, so `reportBot` will conflict and needs to be redone on `context.prisma` after a rebase (inline). The rebase also matters for CI: the green run was against the old base, which doesn't have `schema.graphql` or the Prisma user code, so it doesn't tell us much about main. Two smaller things inline: `reported` can double count under concurrent requests, and `reportBot` returns `true` where legacy returned `{ code, message }`. One question that isn't really yours: legacy stores bot reports as `_reporterId`/`_reportedUserId`, but the new Mongoose and Prisma `BotReport` models use `reporterId`/`userId` with no mapping, so the duplicate check won't see reports made through the live API, and the unique index could clash on those old docs. Fine to ticket that separately unless you want to add the mapping here. Also agree with @flyblackbox on filling out the PR description, please before start working on issues make sure to fetch or pull latest changes, **112 commits behind main creates a big issue**, so next time please fetch latest code to avoid conflict and create unexpected issues, on your next issues do not forget to pull/fetch before start working, thanks @sandy4242 

Address these and this will be ready to be merged.

@flyblackbox @sandy4242

motirebuma:
@sandy4242 can you please address the review gaps.

Thanks @sandy4242

sandy4242:
@motirebuma i am working on it, i will update those once completely done... thanks

